PR #1, opened 2023-03-15
Add initial GitHub Actions workflow

ernestask:


ernestask:
@NordSecurity/libdrop-team

PR #10, opened 2023-04-05
Do not download the file if its already there

matszczygiel:
Added the check if the file to be downloaded is already there.
Needed to change the V3 protocol a bit because previous iteration didn't allow for multiple checksum requests.

There is still a caveat here. I'm not able to resume and check for existence the directory transfer, because it's really hard do deal with it and I don't even know how. I'm leaving the issue behind since with the use of transfers DB the issue might disappear magically

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -1148,7 +1148,7 @@ scenarios = [
         },
     ),
     Scenario(
-        "scenario8",
+        "scenario8-1",
         "Send two identical files one by one, expect no overwrites to happen",
         {
             "ren": ActionList(
@@ -1165,7 +1165,9 @@ scenarios = [
                     ),
                     action.Wait(event.Start(0, "testfile-small")),
                     action.Wait(event.FinishFileUploaded(0, "testfile-small")),
-                    action.NewTransfer("172.20.0.15", ["/tmp/testfile-small"]),
+                    action.NewTransfer(
+                        "172.20.0.15", ["/tmp/duplicate/testfile-small"]
+                    ),
                     action.Wait(
                         event.Queued(
                             1,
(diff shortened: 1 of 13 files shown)


test/scenarios.py
```diff
@@ -1242,9 +1244,144 @@
...
+                    action.NoEvent(),
+                    action.Stop(),
+                ]
+            ),
+        },
+    ),
     Scenario(
         "scenario9",
-        "Send two identical files with extensions one by one, expect no overwrites to happen",
+        "Send the same file twice, expect reporting that the file is already downloaded",
```
LukasPukenis, line 1384: This is nice

LukasPukenis (approved):
I looked into the test cases and they make sense. Haven't looked into the code as it looks heavily refactored and I have little clue what's going on in there.

We test if the file exists and is complete however we do not have tests to exercise the resuming of an incomplete file(or I can't find it in the soup of tests :D).

My idea for that would be that we would need to track Progress events however that's difficult to do deterministically, so an idea would be to allow the test suite to record progress events at checkpoints and compare whether the next progress event was higher or lower. We can do the same with the temporary file size. What do you think?

matszczygiel:
@LukasPukenis yep, that's true that we don't test the resume. Was thinking about introducing a field into `Started` event with the amount already downloaded. Obviously it would by 0 if there is nothing to resume, but that would allow for testing resumes.

Your solution is also an option, but I would rather check the temp file size and then compare with `Progress` event

What do you think ? I think it might be easier to introduce the test case in another PR.

LukasPukenis:
@matszczygiel introducing those tests separately makes perfect sense, I just used this PR as a platform for brainstorming.

While I think introducing the amount inside of `Started` is nice I don't see how it actually helps as it contains the same information as the first `Progress` event. It would be nice though as API users would know that the file is being resumed for almost no cost to us which might be useful for UX purposes.

Can you clarify what you mean by checking the `Progress` event with a temporary file? As I understand such a thing would also test that the progress events properly map to the temp file size which is nice as well

matszczygiel:
@LukasPukenis Actually I thought that the first `Progress` contains the amount the bytes received **before** the first chunk arrives. But I was wrong, it's not the case.
But I think it would be beneficial to make a requirement that the first `Progress` event contains 0 bytes received on fresh file transfer and in case of resume, the amount of bytes already there in the temp file.

If we impose such requirement then we can test if first `Progress.bytes_received` == `temp file size read from disk`
Let me do that in next PR ;)

PR #11, opened 2023-04-06
File resume tests

matszczygiel:
As in the title. I wrote a test for file resuming and one for resuming but with invalid temporary file, to check if this prevents from reusing the file

### DIFF
--- a/drop-transfer/src/protocol/v3.rs
+++ b/drop-transfer/src/protocol/v3.rs
@@ -3,7 +3,7 @@
 //! * client (sender)   -> server (receiver): `TransferRequest`
 //!
 //! If the server has the file or a part of it, the server can request checksum
-//! from the client. In that case sender must reqport the checksum. The request
+//! from the client. In that case sender must report the checksum. The request
 //! can be repeated
 //! * server (receiver) ->   client (sender): `ReqChsum (file)`
 //! * client (sender)   -> server (receiver): `ReportChsum (file)`
--- a/drop-transfer/src/protocol/v3.rs
+++ b/drop-transfer/src/protocol/v3.rs
@@ -3,7 +3,7 @@
 //! * client (sender)   -> server (receiver): `TransferRequest`
 //!
 //! If the server has the file or a part of it, the server can request checksum
-//! from the client. In that case sender mut reqport the checksum. The request
+//! from the client. In that case sender must reqport the checksum. The request
 //! can be repeated
 //! * server (receiver) ->   client (sender): `ReqChsum (file)`
 //! * client (sender)   -> server (receiver): `ReportChsum (file)`
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -201,7 +201,9 @@ class Wait(Action):
         self._event: Event = event
 
     async def run(self, drop: ffi.Drop):
-        await drop._events.wait_for(self._event)
+        await drop._events.wait_for(
+            self._event, not isinstance(self._event, event.Progress)
+        )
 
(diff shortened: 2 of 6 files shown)


LukasPukenis (approved):
Nice test suite! I annotated a few obvious typos.

For completeness' sake, the suite can be improved so that the sender's progress event would be checked as well, however, it doesn't contain a temporary file to check against so unless there's an easy way to do so it's not worth it. Just an idea

test/drop_test/action.py
```diff
@@ -324,3 +326,20 @@ async def run(self, drop: ffi.Drop):
...
+
+    async def run(self, drop: ffi.Drop):
+        stat = os.stat(self._tmp_file_path)
+        await drop._events.wait_for(event.Start(self._uuid_slot, self._file_id), False)
+        await drop._events.wait_for(
+            event.Progress(self._uuid_slot, self._file_id, stat.st_size), False
+        )
+
+    def __str__(self):
+        return f"WaitForResume({print_uuid(self._uuid_slot)}, {self._file_id}, {self._tmp_file_path})"
```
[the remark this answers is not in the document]
matszczygiel: I've tried that but it was simply failing. There is a possibility of race, because if we wait for `Start` and then wait for `Progress` we need to read the file size in between, but the file might be already changed and the `Progress` event already pending before we will wait for it. The solution is to read the file size before `Start`.

matszczygiel:
@LukasPukenis It is possible to check for resume on the sender side since they share the `/tmp` directory. But we currently have some kind of race condition in events dispatch and I found that the first `Progress` event can simply be skipped for sender.
So it's not reliably possible at least before we fix the issue of event races.
Anyway, the check on the receiver is sufficient to say with 100% certainty that the resume happened or not so it's not an issue at this moment. Leaving  this as a little TODO.

drop-transfer/src/protocol/v3.rs
```diff
@@ -3,7 +3,7 @@
 //! * client (sender)   -> server (receiver): `TransferRequest`
 //!
 //! If the server has the file or a part of it, the server can request checksum
-//! from the client. In that case sender mut reqport the checksum. The request
+//! from the client. In that case sender must reqport the checksum. The request
```
Lipt0nas, line 6:
```suggestion
//! from the client. In that case sender must report the checksum. The request
```

test/scenarios.py
```diff
@@ -3145,4 +3145,210 @@
...
+                            },
+                        )
+                    ),
+                    action.Download(
+                        0,
+                        "testfile-big",
+                        "/tmp/received/21-1",
+                    ),
+                    action.Wait(event.Start(0, "testfile-big")),
+                    # wait for the initial progress indicating that we start form the beginning
```
Lipt0nas, line 3207:
```suggestion
                    # wait for the initial progress indicating that we start from the beginning
```

test/scenarios.py
```diff
@@ -3145,4 +3145,210 @@
...
+                            },
+                        )
+                    ),
+                    action.Download(
+                        0,
+                        "testfile-big",
+                        "/tmp/received/21-2",
+                    ),
+                    action.Wait(event.Start(0, "testfile-big")),
+                    # wait for the initial progress indicating that we start form the beginning
```
Lipt0nas, line 3310:
```suggestion
                    # wait for the initial progress indicating that we start from the beginning
```

Lipt0nas (approved):
Some extra small typos

LukasPukenis (dismissed):

PR #12, opened 2023-04-07
Use hex encoding for the file checsum on the wire

matszczygiel:
This change is mainly for aesthetics and compactness. It will also help in case we need to write the messages by hand during testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -233,6 +233,7 @@ dependencies = [
  "drop-config",
  "futures",
  "futures-util",
+ "hex",
  "hyper",
  "infer",
  "libc",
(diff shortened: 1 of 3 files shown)


LukasPukenis:
Without a use case it will look like a random decision, I suggest adding a unit test exercising it

matszczygiel:
@LukasPukenis you mean like serialization tests ? In that case we should include tests for all of the protocol messages

LukasPukenis:
For v3 alone is fine. Without supporting code it might look like a random decision. commit message might suffice but still.

LukasPukenis (approved):

matszczygiel:
Actually I will write the unit tests for messages serialization to ensure no braking changes in the future. Let's use this opportunity to do that, because it's very useful imo

LukasPukenis (approved):

PR #15, opened 2023-04-17
Lock UUIDS global in the test framework

matszczygiel:
This PR attempts to fix spurious race condition in test framework.
The race manifestation example: https://github.com/NordSecurity/libdrop/actions/runs/4720591717/jobs/8372817177?pr=14

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -9,7 +9,7 @@ import typing
 
 from . import event, ffi
 from .logger import logger
-from .event import Event, File, print_uuid, UUIDS
+from .event import Event, File, print_uuid, UUIDS, UUIDS_LOCK
 
 import sys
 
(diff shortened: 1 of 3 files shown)


LukasPukenis:
@matszczygiel the PR looks not offensive but I cant understand what it solves, could you elaborate a bit more?

ernestask:
Makes sense, but I don’t really see how the races can happen.

matszczygiel:
@LukasPukenis sure. When we call `new_transfer(peer, ..)`, we expect to receive the `Queued` event shortly. The test framework matches the uuid to the slot when this happens here https://github.com/NordSecurity/libdrop/blob/main/test/drop_test/ffi.py#L421 The slot itself is registered when calling `NewTransfer` action https://github.com/NordSecurity/libdrop/blob/main/test/drop_test/action.py#L49 just after calling `new_transfer()`
The race I observed in the pipeline linked in the description was that the `Queued` event arrived before python registered the slot when calling `new_transfer()` because events are handled in a different libdrop thread.
I got an error on the line  https://github.com/NordSecurity/libdrop/blob/main/test/drop_test/ffi.py#L421 saying that the uuid does not exist or something.

Locking the `new_transfer()` unlocks the uuid registry only after it successfully registers uuid slot

matszczygiel:
The exact line in the CI job https://github.com/NordSecurity/libdrop/actions/runs/4720591717/jobs/8372817177?pr=14#step:5:471

LukasPukenis:
@matszczygiel thanks, a superb explanation!

PR #16, opened 2023-04-17
Configure Android SONAME

dragan-cecavac-nordsec:
SONAME should be provided by Android libs and is enforced with newer API https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md#missing-soname-enforced-for-api-level-23

Similar change has been introduced to libnudler and is on its way to libtelio.

### DIFF
--- a/norddrop/build.rs
+++ b/norddrop/build.rs
@@ -126,6 +126,11 @@ fn main() -> Result<(), Box<dyn Error>> {
     if target_os == "windows" {
         create_winres(&version)?;
     }
+    if target_os == "android" {
+        let pkg_name = env!("CARGO_PKG_NAME");
+        let soname = format!("lib{}.so", pkg_name);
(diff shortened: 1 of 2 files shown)


norddrop/build.rs
```diff
@@ -126,6 +126,9 @@ fn main() -> Result<(), Box<dyn Error>> {
     if target_os == "windows" {
         create_winres(&version)?;
     }
+    if target_os == "android" {
+        println!("cargo:rustc-cdylib-link-arg=-Wl,-soname,libnorddrop.so");
```
matszczygiel, line 130: Can we construct `libnorddrop.so` from `CARGO_PKG_NAME` env ? In order to protect against renaming etc

matszczygiel (changes requested):

LukasPukenis (approved):
Nice, thanks

matszczygiel (approved):
LGTM

PR #18, opened 2023-04-19
Initial SQLite migration and structure

LukasPukenis:
This PR prepares the groundwork for SQLite3 and is part of upcoming PRs based on this one.


drop-storage/sqlx-data.json
```diff
@@ -0,0 +1,3 @@
+{
```
matszczygiel, line 1: Is this file necessary?

drop-storage/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+
+type Result<T> = std::result::Result<T, Error>;
+// SQLite storage wrapper
+pub struct Storage {
+    _logger: Logger,
+    #[allow(dead_code)]
+    conn: SqlitePool,
+}
+
+#[derive(Debug, strum::Display, strum::EnumString, PartialEq, Eq)]
```
matszczygiel, line 18: `strum` would probably be replaced with something like this https://github.com/launchbadge/sqlx/blob/main/examples/postgres/files/src/main.rs#L4

drop-storage/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        debug!(logger, "SQLite existing: {}", path);
+
+        let mut options = SqliteConnectOptions::from_str(path)?;
+        options.log_statements(log::LevelFilter::Debug);
+
+        let conn = SqlitePool::connect_with(options)
+            .await
+            .map_err(|e| error::Error::InternalError(e.to_string()))
```
matszczygiel, line 93: Duplicated line?

drop-config/src/lib.rs
```diff
@@ -6,13 +6,14 @@ pub struct Config {
     pub moose: MooseConfig,
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
```
matszczygiel, line 9: If we're dropping `Copy` I suggest wrapping config in `Arc` for cheap cloning in `drop-transfer` crate

drop-transfer/src/manager.rs
```diff
@@ -30,9 +30,10 @@ pub struct TransferState {
 
 /// Transfer manager is responsible for keeping track of all ongoing or pending
 /// transfers and their status
-#[derive(Default)]
 pub(crate) struct TransferManager {
     transfers: HashMap<Uuid, TransferState>,
+    #[allow(dead_code)]
+    storage: Arc<Mutex<drop_storage::Storage>>,
```
matszczygiel, line 36: Why do we use `Arc` here?

drop-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "1.0.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+thiserror = "*"
+sqlx = { version = "0.6", features = [ "runtime-tokio-rustls", "sqlite", "time", "chrono" ] }
+slog = { workspace = true }
+tokio = { version = "1", features = ["full"] }
```
matszczygiel, line 12:
```suggestion
tokio = { workspace = true, features = ["full"] }
```
Are you sure that you need all the features?

drop-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+tokio-util = "0.7.3"
+tokio-test = "*"
+strum = { version = "0.24.1", features = ["derive"] }
+inspect = "0.1.2"
+async-std = { version = "1", features = ["attributes", "tokio1"] }
+log = "0.4"
```
matszczygiel, line 18: It's unused, besides we're using `slog` instead of `log` crate

drop-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dependencies]
+thiserror = "*"
+sqlx = { version = "0.6", features = [ "runtime-tokio-rustls", "sqlite", "time", "chrono" ] }
+slog = { workspace = true }
+tokio = { version = "1", features = ["full"] }
+tokio-util = "0.7.3"
+tokio-test = "*"
+strum = { version = "0.24.1", features = ["derive"] }
+inspect = "0.1.2"
```
matszczygiel, line 16: Unused

drop-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+async-std = { version = "1", features = ["attributes", "tokio1"] }
```
matszczygiel, line 17: Do we need this ?

drop-storage/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+thiserror = "*"
+sqlx = { version = "0.6", features = [ "runtime-tokio-rustls", "sqlite", "time", "chrono" ] }
+slog = { workspace = true }
+tokio = { version = "1", features = ["full"] }
+tokio-util = "0.7.3"
+tokio-test = "*"
```
matszczygiel, line 14: Do we need these two?

matszczygiel (changes requested):

drop-storage/migrations/20230406110818_initial_migration.sql
```diff
@@ -0,0 +1,128 @@
...
+  is_outgoing INTEGER NOT NULL,
+
+  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  FOREIGN KEY(peer_id) REFERENCES peers(id) ON DELETE CASCADE ON UPDATE CASCADE,
+  CHECK(is_outgoing = 0 OR is_outgoing = 1)
+);
+
+-- added paths to the outgoing transfer
+CREATE TABLE IF NOT EXISTS outgoing_paths (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
```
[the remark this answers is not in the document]
LukasPukenis: `transfer_id` is used in this table as a foreign key.

drop-storage/sqlx-data.json
LukasPukenis: `sqlx` created it so I assumed it's somewhat needed but I can move it into gitignore I guess

drop-storage/src/lib.rs
LukasPukenis: oh wow, really :D

drop-storage/migrations/20230406110818_initial_migration.sql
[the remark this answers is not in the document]
matszczygiel: Oh, sorry. My bad. I meant the `transfers` table but it's already indexed by uuid. NWM

drop-storage/sqlx-data.json
matszczygiel: Well, it's needed in the offline case. Which probably is irrelevant to the SQLite case. I would go for removing it

drop-storage/Cargo.toml
LukasPukenis: this PR was ripped out of a stable branch, these were used in unit tests, however as this PR doesn't contain them, these are unused, nice catch

drop-storage/sqlx-data.json
LukasPukenis: moved to gitignore and removed it

drop-storage/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            debug!(logger, "SQLite not existing. Creating database at {}", path);
+            Sqlite::create_database(path)
+                .await
+                .map_err(|e| error::Error::InternalError(e.to_string()))?;
+        }
+
+        debug!(logger, "SQLite existing: {}", path);
+
+        let options = SqliteConnectOptions::from_str(path)?;
```
matszczygiel, line 88: The `SqliteConnectOptions` has a `create_if_missing` property. Can we use it to replace this logic?

drop-storage/src/lib.rs
LukasPukenis: That's a nice suggestion, even though the structure is not really used in this MR and thus I totally removed it

drop-storage/src/lib.rs
LukasPukenis: Oh that's nice, thanks

drop-storage/src/lib.rs
matszczygiel: I still see it
image

Is this intentional ?

drop-config/src/lib.rs
LukasPukenis: nice suggestion

drop-transfer/src/transfer.rs
```diff
@@ -16,15 +16,15 @@ pub struct Transfer {
...
+    pub fn new(peer: IpAddr, files: Vec<File>, config: Arc<DropConfig>) -> Result<Self> {
         Self::new_with_uuid(peer, files, Uuid::new_v4(), config)
     }
 
     pub(crate) fn new_with_uuid(
         peer: IpAddr,
         files: Vec<File>,
         uuid: Uuid,
-        config: &DropConfig,
+        config: Arc<DropConfig>,
```
matszczygiel, line 27: Um, why? We don't need ownership of config here, plain reference suffices

norddrop/src/device/mod.rs
```diff
@@ -65,7 +65,7 @@ impl NordDropFFI {
                 cb: event_cb,
                 logger,
             }),
-            config: Config::default(),
+            config: Arc::new(Config::default()),
```
matszczygiel, line 68: Probably storing `Arc` here is a huge pain. I would go for storing here as plain data here, and wrap it into arc inside `Service` structure

matszczygiel (approved):
I think there are many places where we can just pass `&DropConfig`. But this can be sorted out later

drop-transfer/src/transfer.rs
LukasPukenis: I changed DropConfig to be wrapped inside an Arc and fixed a ton of issues surrounding that so it simply slipped here and here. Anyway removed it

norddrop/src/device/mod.rs
LukasPukenis: I think it's not that painful and has certain benefits:
- Easy not to think about lifetimes
- Passing the config and updating it in the future makes it just work

Currently, if I will have it as a plain member and wrap it into an Arc<> when passing, I am gonna pass clones instead thus making each transfer with an explicit clone of a config. This is not an issue and depends on the point of view. I changed it into the non-arc member for the piece of mind

matszczygiel (approved):

norddrop/src/device/mod.rs
```diff
@@ -132,11 +133,20 @@ impl NordDropFFI {
...
+                        ffi::types::NORDDROP_RES_DB_ERROR
+                    })?;
+
             let service = match Service::start(
                 addr,
+                storage,
                 tx,
                 self.logger.clone(),
-                self.config.drop,
+                Arc::new(self.config.clone().drop),
```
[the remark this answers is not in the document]
LukasPukenis: thanks

matszczygiel (approved):
:+1:

PR #19, opened 2023-04-21
Fix bindings, pass IP as string

matszczygiel:
I changed the way we pass the IP address. Now it's passed as a string.
Fixed the bindings


LukasPukenis (approved):
This looks like it contains some handwritten Java code which I assume is not really tested. I suggest contacting the Android team and asking them to grab this build and try to exercise that callback.

Because this unblocks the pipeline it makes sense to land it anyway

PR #20, opened 2023-04-24
fix not receiving `FileUploadStarted` and `FileUploadSuccess` events …

matszczygiel:
…before the transfer is canceled

As it turned out, the `FileUploadStart` event (which blocks all other events) was emitted in the async thread, so for small files sometimes it lagged after the transfer cancelation. This resulted in no events being emitted about a particular file upload, despite it being fully transferred.

Emitting the `FileUploadStart` on the same thread as WS message handling should fix the problem.

I managed to reproduce the problem with `udrop`. After applying the fix, the problem seems to be gone

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+### UNRELEASED
+### **UNRELEASED**
+---
+* Fix not receiving `FileUploadStarted` and `FileUploadSuccess` events before the transfer is canceled
+
+---
+<br>
+
 ### v2.0.0
 ### **Brzęczyszczykiewicz**
 ---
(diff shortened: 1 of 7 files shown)


LukasPukenis (approved):
Explanation and implementation makes sense.

IMHO This should be easily reproducible through the tests if removed the `NetworkConfiguration` step to gimp the network. What do you think?

matszczygiel:
@LukasPukenis Let me give it a try ;)

matszczygiel:
@LukasPukenis you were right. The newly added test passes cleanly but fails on the `main` release build. Thus "proving" that this PR fixes the problem

PR #23, opened 2023-04-25
Canceled events has flipped by peer field

matszczygiel:
When canceling file upload from the receiver the events have the wrong `by_peer` value


drop-transfer/src/ws/server/v3.rs
```diff
@@ -149,7 +149,7 @@ impl HandlerLoop<'_> {
         let msg = v3::ServerMsg::Cancel(v3::Cancel { file: file.clone() });
         socket.send(Message::from(&msg)).await?;
 
-        self.on_cancel(file).await;
```
LukasPukenis, line 152: This comment refers to all client/server implementations of v[1|2|3].rs files :)
I find it a bit confusing to understand the client and server relationship when reading the code, for example:
```
// client/v3.rs
async fn on_req(&mut self, socket: &mut WebSocket, req: ClientReq) -> anyhow::Result<()> {
        match req {
            ClientReq::Cancel { file } => self.issue_cancel(socket, file).await,
        }
    }
```

```
// server/v3.rs
async fn on_req(&mut self, ws: &mut WebSocket, req: ServerReq) -> anyhow::Result<()> {
        match req {
            ServerReq::Download { task } => self.issue_download(ws, *task)?,
            ServerReq::Cancel { file } => self.issue_cancel(ws, file).await?,
        }

        Ok(())
    }
```

So both the client and the server can process requests and the client and server means the sender and the receiver appropriately? That makes sense. But then:
```
// client/v3.rs
async fn issue_cancel(&mut self, socket: &mut WebSocket, file: FileId) -> anyhow::Result<()> {
        let msg = v3::ClientMsg::Cancel(v3::Cancel { file: file.clone() });
        socket.send(Message::from(&msg)).await?;

        self.on_cancel(file, false).await;

        Ok(())
    }
```

```
// v3/server.rs
 async fn issue_cancel(&mut self, socket: &mut WebSocket, file: FileId) -> anyhow::Result<()> {
        debug!(self.logger, "ServerHandler::issue_cancel");

        let msg = v3::ServerMsg::Cancel(v3::Cancel { file: file.clone() });
        socket.send(Message::from(&msg)).await?;

        self.on_cancel(file, false).await;

        Ok(())
    }
```

The code always passed `false` into `by_peer` flag which should be the opposite - if we're handling a message from another peer it's should be `true` always and `false` when we're the invoker.

matszczygiel: The "request" in this context is a request coming from the API layer, not the WebSocket. Hence the names `issue_...`.
For example, when the receiver app calls `download_file()` then in `server/v3.rs` the method `on_req()` is fired resulting in calling `server/v3.rs` `issue_download()`. The same goes for canceling. 
Note that in `issue_cancled()` there is a packet "Cancel" send down the WebSocket to notify another peer about the cancelation

matszczygiel: The corresponding method called on WebSocket message is `on_recv()`

LukasPukenis: It's really unexpected to be the caller and server/* code to be called.. then it propagates to the other end and what code is called in there then?

Would there be some way to "split" it so the caller code would map into the `client` and the `sever` code would only be called on other side's requests?

matszczygiel: OK, let's state that the sender is always a client and the receiver always a server from the WebSocket point of view.

So the flow is in case the sender cancels the file:
* senders `cancel_file()` method is called by the app
* senders `on_req()` gets called with `CanceReq` in effect so it enters the `issue_cancel()` method
* sender sends a `Canceled` message down the WebSocket and emits a `Canceled` event with `by_peer = false`
* receiver gets WS message `Canceled` as it enters the `on_cancel()` method
* receiver emits `Canceled` event with `by_peer = true`

I don't think that we can map `client` and `server` because both of them can be an initiator of canceling. The meaning of `server` and `client` here is closely tied to how WebSocket works

LukasPukenis (approved):
No reason to block this MR especially when it's backed by a testcase. Let's have a discussion some time later about the architecture

PR #25, opened 2023-04-27
Moose init error only on non-prod

Lipt0nas:
Return moose error only when on non-prod, otherwise just use the mock implementation

### DIFF
--- a/norddrop/src/device/mod.rs
+++ b/norddrop/src/device/mod.rs
@@ -104,7 +104,12 @@ impl NordDropFFI {
             Ok(moose) => moose,
             Err(err) => {
                 error!(logger, "Failed to init moose: {:?}", err);
-                return Err(ffi::types::NORDDROP_RES_ERROR);
+
+                if self.config.moose.prod == false {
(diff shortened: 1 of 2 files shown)


LukasPukenis (approved):
a really elegant solution it looks like

norddrop/src/device/mod.rs
```diff
@@ -104,7 +104,12 @@ impl NordDropFFI {
             Ok(moose) => moose,
             Err(err) => {
                 error!(logger, "Failed to init moose: {:?}", err);
-                return Err(ffi::types::NORDDROP_RES_ERROR);
+
+                if !self.config.moose.prod {
+                    return Err(ffi::types::NORDDROP_RES_ERROR);
+                }
+
```
matszczygiel, line 111: I would log a warning here: "Using moose mock" or something

matszczygiel (approved):

norddrop/src/device/mod.rs
Lipt0nas: good point :+1:

matszczygiel (approved):

PR #26, opened 2023-04-27
norddrop: Add lib crate type back

ernestask:
Since the platforms builds now specify their own crate types in the command line, we can bring back the lib type for Linux builds.


LukasPukenis (approved):
Assuming windows builds properly yay!

PR #28, opened 2023-05-08
Fix C# pubkey callback bindings

LukasPukenis:
Current declaration was
```
out byte[]
```
Which in C# meant that it must be overwritten
The proper fix is to have
```
byte*
```

This is not as nice however it solves the problem as C# now is able to directly write data to this buffer


Lipt0nas (approved):
Looks good :+1:

PR #29, opened 2023-05-08
Shared crate for types

Lipt0nas:
It's becoming more and more painful to share common types between all of the crates we have, so before the circular dependencies get out of hand, we should move everything out to a separate crate

### DIFF
--- a/drop-transfer/Cargo.toml
+++ b/drop-transfer/Cargo.toml
@@ -19,7 +19,7 @@ tempfile = "3.5.0"
 anyhow = "1.0.70"
 async-trait = { workspace = true }
 drop-analytics = { version = "1.0.0", path = "../drop-analytics" }
-drop-config = { version = "1.0.0", path = "../drop-config" }
+drop-types = { version = "1.0.0", path = "../drop-types" }
 drop-auth = { path = "../drop-auth" }
 drop-storage = { version = "1.0.0", path = "../drop-storage" }
 futures = "0.3"
--- a/drop-types/src/ffi.rs
+++ b/drop-types/src/ffi.rs
@@ -1,0 +1,113 @@
+use std::time::Duration;
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct Config {
+    pub dir_depth_limit: usize,
+    pub transfer_file_limit: usize,
+    pub req_connection_timeout_ms: u64,
+    #[serde(default = "default_connection_max_retry_interval_ms")]
+    pub connection_max_retry_interval_ms: u64,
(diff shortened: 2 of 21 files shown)


drop-transfer/Cargo.toml
```diff
@@ -19,7 +19,7 @@ tempfile = "3.5.0"
 anyhow = "1.0.70"
 async-trait = { workspace = true }
 drop-analytics = { version = "1.0.0", path = "../drop-analytics" }
-drop-config = { version = "1.0.0", path = "../drop-config" }
+drop-types = { version = "1.0.0", path = "../drop-types" }
```
matszczygiel, line 22: I think it's better to remove `version = "1.0.0"` and let the cargo pick whatever is in the `../drop-types`

drop-types/src/ffi.rs
```diff
@@ -0,0 +1,113 @@
+use std::time::Duration;
+
+use serde::Deserialize;
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct Config {
```
matszczygiel, line 6: I don't like the idea of `ffi` module here and perhaps leave all ffi related stuff in `norddrop` crate. But if you insist, let's move it out of this module and rename it to `ConfigJson` or something like that, because the purpose of this structure is to define the JSON representation of the actual config.

matszczygiel (approved):
I like the organize things idea :+1:

matszczygiel:
Oh, I'm sorry, didn't notice it's in the draft  stage :laughing:

LukasPukenis:
The description is about circular dependencies but in reality - it just indicates that there is a wrong separation of concerns in the code. Moving types into a commonplace might work if they are truly meant to be common but from this MR what I can see:
- FFI types should only be known in the FFI layer
- Config should not be known by everyone but only the first layer that accepts it really. For example `File` only uses `dir_depth_limit` and  `transfer_file_limit` from it and should have a structure of its own to accept what it needs or just straight parameters.

Verbosity is better in the long run from my experience than just applying DRY blindly just because the code repeats. The code might look the same but the semantic meaning of it is totally different. For example, a `Transfer` in `drop-transfer` encapsulates a transfer and its behavior but in `drop-storage` it encapsulates a wrapper around the database type. They have nothing in common actually just the same name and a few fields coincidentally like UUID. We can move the Transfer as a POD into a common crate but then it will contain fields that are irrelevant for one or another crate so not much meaning is introduced there.

What I would really suggest is isolating each crate as much as possible and making only thin connections between these crates/modules. This way we can reason about stuff easily as well as make changes. As a side effect, this will increase the number of structs, crates, etc we have but will make the project more manageable and understandable.

PR #32, opened 2023-05-11
Implement file id as hash

matszczygiel:
I've implemented the file identification by hashing the filesystem paths. It is used in events and the API layer.
For compatibility and file placement reasons the old Id still is in use, but has been renamed to `FileSubPath`.

Test cases are also adjusted but since the receiving peer does not have access to the full path of the file, it is forced to use the subpath as an ID. The implementation allows for that because `FileId` is just a `String` under the hood.

Also the `RequestQueued/Received` now contains the flat file list, with Id and subpath as separate fields

### DIFF
--- a/drop-transfer/src/ws/client/v3.rs
+++ b/drop-transfer/src/ws/client/v3.rs
@@ -12,7 +12,7 @@ use tokio::{sync::mpsc::Sender, task::JoinHandle};
 use tokio_tungstenite::tungstenite::{self, Message};
 
 use super::{handler, ClientReq, WebSocket};
-use crate::{protocol::v3, service::State, utils::Hidden, ws, FileSubPath};
+use crate::{file::FileSubPath, protocol::v3, service::State, utils::Hidden, ws};
 
 pub struct HandlerInit<'a> {
     state: Arc<State>,
@@ -95,21 +95,23 @@ impl HandlerLoop<'_> {
             if !task.job.is_finished() {
                 task.job.abort();
 
+                let file = self
+                    .xfer
+                    .file_by_subpath(&file)
+                    .expect("File should exists since we have a transfer task running");
+
                 self.state.moose.service_quality_transfer_file(
                     Err(u32::from(&crate::Error::Canceled) as i32),
                     drop_analytics::Phase::End,
                     self.xfer.id().to_string(),
                     0,
-                    self.xfer
-                        .file_by_subpath(&file)
-                        .expect("File should exists since we have a transfer task running")
-                        .info(),
+                    file.info(),
                 );
 
                 task.events
                     .stop(crate::Event::FileUploadCancelled(
                         self.xfer.clone(),
-                        file,
+                        file.id().clone(),
                         by_peer,
                     ))
                     .await;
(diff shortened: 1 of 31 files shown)


drop-transfer/src/ws/client/v3.rs
```diff
@@ -152,7 +152,10 @@ impl HandlerLoop<'_> {
         limit: u64,
     ) -> anyhow::Result<()> {
         let f = || {
-            let xfile = self.xfer.file(&file_id).context("File not found")?;
+            let xfile = self
+                .xfer
+                .file_by_subpath(&file_id)
```
LukasPukenis, line 180: Could you clarify if I understand it correctly as it's a lot to sift through:

Transfer now holds the files as a flat list instead of a tree hierarchy and thus has `file_by_subpath()`. This sounds a bit weird as it only makes sense for old versions as new ones will use the hash instead, but we have a FileId conversion to a FileSubpath so as both are strings it just works.

Is it correct that the old versions will pass the glued directory as an ID and also will expect the same while this version will pass and expect a hash instead? This way we're actually breaking the compatibility between this release and others with no indication as it will just respond that such a file was not found.

If what's above is true we then should:
- introduce additional meta-data for `file_by_subpath` so it would resolve either file IDs or hashes

Optionally introduce v4 to not complexify v3 implementation.

test/drop_test/config.py
```diff
@@ -24,3 +24,84 @@ def __init__(self, ip: str, pubkey: bytes, privkey: bytes):
     ),
 }
 # fmt: on
+
+
+class File:
+    def __init__(self, size: int, id: str):
+        # in kilobytes
+        self.size = size
+        self.id = id
```
LukasPukenis, line 33: I was thinking about reusing the `File` from `Action` but this actually is pretty nice. It's just a bit confusing as we have 3 classes called `File` in the test framework: Event, Action, Config. Maybe it makes sense to rename on of them?

LukasPukenis (commented):
This is initial review, I will continue the next day

norddrop/src/device/types.rs
```diff
@@ -220,11 +207,47 @@ impl From<drop_transfer::Transfer> for EventRequestQueued {
...
+
+    for tr_file in t.files().values() {
+        let mut iter = tr_file.subpath().iter();
+        let name = if let Some(name) = iter.next_back() {
+            name.clone()
+        } else {
+            continue;
+        };
+
+        let mut files = &mut files;
```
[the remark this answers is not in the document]
matszczygiel: The function is no longer needed after deciding to use a flat file list in events

norddrop/src/device/types.rs
```diff
@@ -220,11 +207,47 @@ impl From<drop_transfer::Transfer> for EventRequestQueued {
...
         EventRequestQueued {
             transfer: t.id().to_string(),
-            files: t.files().iter().map(|(_, v)| v.into()).collect(),
+            files: extract_transfer_files(&t).into_values().collect(),
         }
     }
 }
 
+fn extract_transfer_files(t: &drop_transfer::Transfer) -> HashMap<String, File> {
+    let mut files: HashMap<String, File> = HashMap::new();
```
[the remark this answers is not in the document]
matszczygiel: Same, this function is deleted by later commits

drop-transfer/src/ws/client/v3.rs
matszczygiel: You're somewhat right. The files as indexed by hash. However old protocols v1..=v3 operate on subpaths, so for this particular `client/v3.rs` implementation we still need to query the files by subpath.
In the next PR I'm going to make a v4 protocol, operating on hashes exclusively. I didn't include it in this PR because it would be humongous.
So this PR changes the subpaths -> hashes in the internals, API layer, and events. The next one will make use of it for the v4 protocol to properly fix the name-clash bugs

test/drop_test/config.py
matszczygiel: Changed `config.File` -> `config.TestFile`

LukasPukenis (dismissed):

PR #33, opened 2023-05-11
Persist transfers

Lipt0nas:


drop-storage/.env
```diff
@@ -1 +1 @@
-DATABASE_URL="sqlite://testdb.sqlite3"
+DATABASE_URL="sqlite://drop-storage/testdb.sqlite3"
```
matszczygiel, line 1: I would change the file name to `.env.example` or something like that so that it won't create sqlite files somewhere accidentally

drop-storage/Cargo.toml
```diff
@@ -9,3 +9,4 @@ edition = "2021"
 thiserror = "*"
 sqlx = { version = "0.6", features = [ "runtime-tokio-rustls", "sqlite", "time", "chrono" ] }
 slog = { workspace = true }
+serde = { version = "1.0.160" }
```
matszczygiel, line 12: It's used in at least 2 other crates. I'm suggesting moving it to workspace `Cargo.toml` to avoid duplicate versions

drop-storage/migrations/20230406110818_initial_migration.sql
```diff
@@ -1,127 +1,142 @@
...
 
 -- states for outgoing paths(uploads)
-CREATE TABLE IF NOT EXISTS outgoing_traversed_path_pending_states (
+CREATE TABLE IF NOT EXISTS outgoing_path_pending_states (
   id INTEGER PRIMARY KEY AUTOINCREMENT, 
-  path_id TEXT NOT NULL,
-  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, 
-  FOREIGN KEY(path_id) REFERENCES outgoing_traversed_paths(id) ON DELETE CASCADE ON UPDATE CASCADE
+  path_id INTEGER NOT NULL,
+  created_at TIMESTAMP NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
```
matszczygiel, line 66: I suggest using a different name. `path_id` is used in the table above but it's a `TEXT`. Where is an `INTEGER`

drop-storage/src/lib.rs
```diff
@@ -32,4 +35,671 @@ impl Storage {
...
+        .await
+        .map_err(error::Error::DBError)
+        .map(|_| ())
+    }
+
+    pub async fn purge_transfers(&self, transfer_ids: Vec<String>) -> Result<()> {
+        let mut conn = self.conn.acquire().await?;
+
+        let transfer_ids = transfer_ids.join(",");
+        sqlx::query!("DELETE FROM transfers WHERE id IN (?1)", transfer_ids)
```
matszczygiel, line 396: Is it possible to pass the array into the query directly?

drop-storage/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+    FileCanceled(TransferType, TransferId, FilePath, bool),
+    TransferCanceled(TransferType, TransferInfo, bool),
+
+    FileFailed(TransferType, TransferId, FilePath, u32),
+    TransferFailed(TransferType, TransferInfo, u32),
+
+    FileUploadComplete(TransferId, FilePath),
+    FileDownloadComplete(TransferId, FilePath, String),
+
+    Progress(TransferId, FilePath, i64),
```
matszczygiel, line 49: Since we're writing the events from scratch, I would suggest not using things like `bool`, `u32`. It's way too confusing. We should either make named fields inside the enum variants or use a newtypes

matszczygiel:
It's nice in general, I have small nitpicks but I have a bigger comment:
* I really do not see the point of the `storage::Event` existence. It just adds one level of indirection which make the whole thing harder to comprehend IMHO
* I like the design of DB, it's well-normalized but it's not always good. Imagine we store the stages as a JSON in a single array. I know that's not "professional" but I think it may actually simplify things which has value in itself

Threat all my comments as suggestions, not as blockers or something like that :)

matszczygiel (dismissed):

drop-storage/src/lib.rs
```diff
@@ -32,4 +35,671 @@ impl Storage {
             conn,
         })
     }
+
+    pub async fn insert_transfer(
```
LukasPukenis, line 39: The queries will be executed one by one and will cause a race behavior. For example, the paths might not be yet added completely so the API might return an incomplete state.

To fight this databases have this concept - transaction. https://docs.rs/sqlx/latest/sqlx/struct.Transaction.html sqlx seems to support it. So basically all queries that should be atomic, should be wrapped in a transaction.

We might not encounter an issue as we lock the drop-storage instance in a mutex and only expect our own app to read it. Still leaving the comment just to be aware

drop-storage/src/lib.rs
```diff
@@ -32,4 +35,671 @@ impl Storage {
...
+                path_id: s.path_id,
+                created_at: s.created_at.timestamp_millis(),
+            })
+            .collect();
+        }
+
+        Ok(paths)
+    }
+
+    async fn get_incoming_paths(&self, transfer_id: String) -> Result<Vec<IncomingPath>> {
```
LukasPukenis, line 600: This whole file really begs for unit tests. it would be great to simulate insertions of a few transfers with few files, imitate several fails, cancels and etc(really all the calls) and except the proper data to be returned.

drop-transfer/src/event.rs
```diff
@@ -28,7 +28,7 @@ pub enum Event {
     FileUploadFailed(Transfer, FileId, Error),
     FileDownloadFailed(Transfer, FileId, Error),
 
-    TransferCanceled(Transfer, bool),
+    TransferCanceled(Transfer, bool, bool),
```
LukasPukenis, line 31: two bools in a row is a very error-prone way. Let's introduce a struct here and use it as a second parameter instead of two bools

drop-transfer/src/event.rs
```diff
@@ -28,7 +28,7 @@ pub enum Event {
...
 
-    TransferFailed(Transfer, Error),
+    TransferFailed(Transfer, Error, bool),
```
LukasPukenis, line 33: Bool is not a very descriptive name as it depends on the context. For such case I would recommend introducing an enum as it reduces guesswork

drop-storage/.env
Lipt0nas: This file is used to tell sqlx where the database is so it could verify the `sqlx::query!` calls, without it the build would fail, unless we define the variable in the build command, and it also messes up linting in IDE's because of the same problem

drop-storage/src/types.rs
LukasPukenis, line 49: `usize` is more suitable for progress as it can never be negative

LukasPukenis (approved):
Overall really nice. Hard to say about the correctness but it looks very reasonable. Left few minor comments but those can be merged in separately.

drop-storage/.env
matszczygiel: A better solution is to remove `.env` and `testdb.sqlite` and use the offline mode of sqlx https://github.com/launchbadge/sqlx/tree/main/sqlx-cli#enable-building-in-offline-mode-with-query

drop-storage/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+impl From<TransferType> for i32 {
+    fn from(transfer_type: TransferType) -> Self {
+        match transfer_type {
+            TransferType::Incoming => 0,
+            TransferType::Outgoing => 1,
+        }
+    }
+}
```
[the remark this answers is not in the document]
LukasPukenis: This is very explicit and clear. What happens if I sometime later switch the places of members by using repr?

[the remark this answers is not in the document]
matszczygiel: It doesn't change because the enum variants already have the explicit discriminants

drop-storage/src/lib.rs
Lipt0nas: sadly sqlx currently doesn't support this

drop-storage/src/types.rs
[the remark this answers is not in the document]
LukasPukenis: If we have
```
#[repr(u32)]
enum Transfer {
    Outgoing,
    Incoming,
}

```
then it will be converted to a different number given that we switch `Outgoing` and `Incoming`. With the current implementation, we're set in stone to return 0 and 1 explicitly. Also helps with readability a bit.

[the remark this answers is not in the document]
matszczygiel: So let's use:
```rust
#[repr(u32)]
enum Transfer {
    Incoming = 0,
    Outgoing = 1,
}
```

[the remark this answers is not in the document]
LukasPukenis: That's nice as well

drop-transfer/src/event.rs
Lipt0nas: done

drop-transfer/src/event.rs
Lipt0nas: fixed

drop-storage/src/types.rs
Lipt0nas: fixed

drop-storage/src/types.rs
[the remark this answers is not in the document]
Lipt0nas: done

drop-storage/src/lib.rs
```diff
@@ -32,4 +35,671 @@ impl Storage {
...
+        for path in transfer.files {
+            self.insert_path(transfer_type, transfer.id.clone(), path, &mut conn)
+                .await?;
+        }
+
+        Ok(())
+    }
+
+    async fn insert_path(
+        &self,
```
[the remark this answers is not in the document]
Lipt0nas: fixed

drop-storage/migrations/20230406110818_initial_migration.sql
```diff
@@ -1,127 +1,142 @@
 CREATE TABLE IF NOT EXISTS peers (
-  id TEXT PRIMARY KEY,
-  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
```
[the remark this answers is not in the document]
Lipt0nas: As discussed, we would still need to be writing additional non-default query code to make this work, retrieving this variable from the database is trivial due to `sqlx`, and we get the added benefit of working with an actual time variable inside the database if need be in this case

drop-storage/migrations/20230406110818_initial_migration.sql
Lipt0nas: renamed `path_id` to `path_hash` to better convey the value and to keep the naming in line with `transfer_id`

drop-storage/migrations/20230406110818_initial_migration.sql
```diff
@@ -1,127 +1,142 @@
...
 
--- traversed paths for outgoing transfer at the time of the transfer creation
--- this is used to "freeze" the state of the files so we could choose to either traverse again or use the traversed paths as origin for the transfer
-CREATE TABLE IF NOT EXISTS outgoing_traversed_paths (
-  id INTEGER PRIMARY KEY AUTOINCREMENT, 
-  path_id INTEGER NOT NULL, 
+CREATE TABLE IF NOT EXISTS transfer_cancel_states (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  transfer_id TEXT NOT NULL,
+  by_peer INTEGER NOT NULL,
```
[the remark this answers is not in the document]
Lipt0nas: added

drop-storage/migrations/20230406110818_initial_migration.sql
```diff
@@ -1,127 +1,142 @@
...
+  FOREIGN KEY(transfer_id) REFERENCES transfers(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+-- all the paths inside the outgoing transfer
+CREATE TABLE IF NOT EXISTS outgoing_paths (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  transfer_id TEXT NOT NULL,
+  path TEXT NOT NULL,
+  path_id TEXT NOT NULL,
   bytes INT NOT NULL,
```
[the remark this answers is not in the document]
Lipt0nas: added

drop-storage/src/lib.rs
Lipt0nas: Created a ticket for this

drop-storage/src/lib.rs
Lipt0nas: done

drop-storage/src/types.rs
Lipt0nas: As discussed, leaving this information for the user to keep in mind a better

drop-storage/.env
Lipt0nas: changed it to offline mode, one additional idea might be to add the `sqlite` and `env` files to `gitignore` so that locally you could use online mode, since it takes precedence

matszczygiel: Um, it's not what I meant by offline mode. My point was to not include the `offline` feature but to run `cargo sqlx prepare` then it generates the `<something>.sqlx` file which is referred to be the macros during the compilation. The `DATABASE_URL` still takes precedence but it's no longer needed if you're compiling it on CI or do not care about SQL development at this moment.
Regarding the `.env` it's a good idea to add it to `.gitignore` :+1:

Lipt0nas: I've updated the `.gitignore` file

As for the `offline` feature, this is actually needed to generate and parse the `****.sqlx` file, `sqlx` complained about it during the generation, and it makes sense if you look at the diff in the `Cargo.lock` file, as it pulls in `serde` dependencies

matszczygiel: Are you sure? This is from sqlx README
image

Lipt0nas: This is the current README for the upcoming `0.7.0` version, if we look at the readme from the `0.6` version that we are using
image

source: https://github.com/launchbadge/sqlx/blob/17cebde64aa6313d8c3efe2db6a05bdcff48def0/README.md

matszczygiel: Ok, my bad :facepalm:

changelog.md
```diff
@@ -5,6 +5,7 @@
 * File IDs are no longer valid paths. Changed the structure of `RequestQueued/Received` events to contain a flat file list
 * Fix duplicate filenames error
 * Removed libnorddrop 2.0.0 wire protocol, because of the security issue
+* Persist transfers to SQLite
```
matszczygiel, line 8: This one is for v4.0.0

Lipt0nas: my bad, good catch :+1:

PR #34, opened 2023-05-16
Lukasp/file 304 chacha to hmacs

LukasPukenis:


drop-auth/src/lib.rs
```diff
@@ -14,6 +14,39 @@ pub const SECRET_KEY_LENGTH: usize = 32;
...
+}
+
+impl Deref for SecretKey {
+    type Target = x25519_dalek::StaticSecret;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
```
matszczygiel, line 49:
```suggestion
pub use x25519_dalek::PublicKey;
pub use x25519_dalek::StaticSecret as SecretKey;
```
A bit simpler

matszczygiel (approved):

matszczygiel (approved):

PR #35, opened 2023-05-17
V4 protocol

matszczygiel:
Implementation of V4 protocol. The difference between V3 is the file ID sync


LukasPukenis (approved):
I see this as a normal continuation of the previous PR so lets just land it

PR #37, opened 2023-05-17
Sending multiple files with the same filename sends only one

matszczygiel:
This PR has the following changes:

* change the temp file naming scheme, to avoid conflicts on duplicate names
* test cases for duplicate file names
* making temp -> target rename more atomic

These changes guarantee proper handling of duplicated file names as well as Windows names collision due to the case insensitivity

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -3608,7 +3608,7 @@ scenarios = [
                     action.WaitForResume(
                         1,
                         FILES["testfile-big"].id,
-                        "/tmp/received/21-1/testfile-big.dropdl-part",
+                        f"/tmp/received/21-1/{FILES['testfile-big'].id}.dropdl-part",
                     ),
                     action.Wait(
                         event.FinishFileDownloaded(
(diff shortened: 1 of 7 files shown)


test/scenarios.py
```diff
@@ -3810,4 +3812,222 @@
...
+            "stimpy": ActionList(
+                [
+                    action.Wait(
+                        event.Receive(
+                            0,
+                            "172.20.0.5",
+                            {
+                                event.File(
+                                    FILES["testfile-small"].id,
+                                    "testfile-small",
```
LukasPukenis, line 3871: I don't think `event.File` needs both ID fields anymore as only the first ID is enough

LukasPukenis: Hmm,, you're right. Maybe putting those filenames in the `FILES` would make it more clear?

matszczygiel: Um still not true. For example, if you have a file `/tmp/foo/bar` then you have two options: either someone sends you `/tmp/foo` or `/tmp/foo/bar`. And you need to know that in the first case, you save the file as `foo/bar` and in the second case as `bar` in the destination directory.
Both of these cases are exercised in the integration tests, so you cannot assume in advance the file subpath used on a particular file in `config.py`

LukasPukenis (approved):

PR #38, opened 2023-05-17
Lukasp/test ci lockfile

LukasPukenis:


matszczygiel:
@LukasPukenis `chrono` 0.4.23 and 0.4.24 have exactly the same `wasm-bindgen` version specified. There is no point in downgrading it

PR #39, opened 2023-05-18
.github: workflows: lint: Check out HEAD of PR target

ernestask:
Currently, the checkout action checks out a merge commit, linting code that theoretically does not exist. This commit changes the ref to be the target of the PR.


ernestask:
@Lipt0nas, try this.

PR #41, opened 2023-05-18
File iIDs don't work on android

matszczygiel:


drop-transfer/src/file/mod.rs
```diff
@@ -148,7 +134,8 @@ impl File {
     }
 
     #[cfg(unix)]
-    fn new_with_fd(subpath: FileSubPath, path: &Path, fd: RawFd) -> Result<Self, Error> {
+    pub fn from_fd(path: impl AsRef<Path>, fd: RawFd, unique_id: usize) -> Result<Self, Error> {
```
LukasPukenis, line 137: I think this can be simplified a bit - this still has a possibility to clash if we would provide the same FD at the same index. Instead of some atomic counter, I would suggest using UUID as the "random" element just to rule out the possibility. It then allows us to hash based on that ID alone instead of path and other elements.

Or atomic counter

LukasPukenis (approved):

drop-transfer/src/file/mod.rs
matszczygiel: I mean, we control the `unique_id` and I've ensured it's unique within a single transfer.
I do see that it needs to be consistent in order to hardcore ID in the integration tests so atomic counter for example is not an option as well as other sources of "randomness" like UUID

LukasPukenis: UUIDs don't really work nicely with determinism, but a unique ID alongside the path doesn't prevent the collision perfectly. I can have two transfers with one(same) file each and there's a slight chance the FDs will be the same.

Honestly probably too little of an issue to have a headache but having `unique_id` isn't really enforced in any way except in the name

LukasPukenis: My bad, it's really OK to have the same ID across the transfers, sorry.

LukasPukenis (approved):

PR #42, opened 2023-05-18
Remove V3 protocol since its yanked

matszczygiel:


LukasPukenis (approved):
less code == neat

LukasPukenis (approved):
less code == neat

PR #45, opened 2023-05-24
Fix go bindings for pubkey callback

matszczygiel:
There was a typo in the `norddropgo.i` swig file


LukasPukenis (approved):
wow, that's a good catch

PR #47, opened 2023-05-24
Remove check for already downloaded files

matszczygiel:
This PR removes the check for already downloaded files. This feature was introduced along the resumes but turned out to be a bug

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,12 @@
+### UNRELEASE
+### **Calm**
+---
+* Remove reporting already downloaded file as finished. Instead, redownload it with (1) suffix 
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -1,3 +1,12 @@
+### UNRELEASE
+### **ANGRY**
```
LukasPukenis, line 2: ANGERY again? :D

LukasPukenis (approved):

changelog.md
matszczygiel: Let me change to `CALM`

LukasPukenis (approved):

PR #48, opened 2023-05-25
Parse file subpath from the relative path

matszczygiel:
This is a fix for Windows issues when sending the directories. The `FileSubPath` was parsed incorrectly from the filesystem path

### DIFF
--- a/drop-transfer/src/file/id.rs
+++ b/drop-transfer/src/file/id.rs
@@ -92,6 +92,15 @@ impl FileSubPath {
     pub fn name(&self) -> &str {
         self.0.last().expect("Missing last path component")
     }
+
+    pub fn from_path(path: impl AsRef<Path>) -> crate::Result<Self> {
+        let vec = path
+            .as_ref()
+            .iter()
+            .map(|cmp| cmp.to_str().map(String::from).ok_or(crate::Error::BadPath))
+            .collect::<Result<_, _>>()?;
+        Ok(Self(vec))
+    }
 }
 
 impl<T> From<T> for FileSubPath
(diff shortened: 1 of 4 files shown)


norddrop/src/device/mod.rs
```diff
@@ -406,8 +406,8 @@ fn prepare_transfer_files(
 ) -> Result<Vec<File>> {
     let mut files = Vec::new();
 
+    #[allow(unused_variables)]
```
LukasPukenis, line 409: why was this moved?

drop-transfer/src/file/id.rs
```diff
@@ -92,6 +92,15 @@ impl FileSubPath {
...
     }
+
+    pub fn from_path(path: impl AsRef<Path>) -> crate::Result<Self> {
+        let vec = path
+            .as_ref()
+            .iter()
+            .map(|cmp| cmp.to_str().map(String::from).ok_or(crate::Error::BadPath))
+            .collect::<Result<_, _>>()?;
+        Ok(Self(vec))
+    }
```
LukasPukenis, line 103: Could you provide more context on what this does and what was wrong with the previous approach?

norddrop/src/device/mod.rs
matszczygiel: Because of `i` in the for loop, it's unused on Windows and generated a warning without this change

drop-transfer/src/file/id.rs
matszczygiel: This one constructs the `FileSubPath` from path components. Previously we only had a method that constructed it by splitting the path string on `/` which is clearly incorrect for Windows.
I mean both methods are needed because the string representation used in the WS protocol is always using `/` but when constructing from the actual filesystem path we need to use this method

LukasPukenis: Thanks for explaining, make sense

LukasPukenis (approved):

PR #50, opened 2023-05-25
Improve C# bindings

matszczygiel:
To be honest it's the only thing we can reasonably do with the C# bindings without writing the wrapper manually

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 * Remove reporting already downloaded file as finished. Instead, redownload it with (1) suffix 
 * Fix the go bindings being unusable
 * Fix directory flattening on Windows
+* Use `byte[]` for the private key in C# bindings
 
 ---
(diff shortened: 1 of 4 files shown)


LukasPukenis:
Why does changing the private key from a string to a pointer fix the bindings? A string should actually be just fine IMHO :)

matszczygiel:
It's in the changes suggested in `NordDrop.cs`, there's more but I cannot find a way to easily implement those

LukasPukenis:
Was this built and checked with Windows to see if it's really enough then? It looks weird. The only change here that could potentially fix something IMHO is `unsafe` keyword added to the callback

matszczygiel:
I've asked the Windows guys to see if the changes in this PR suit them. Otherwise, we can just close it

LukasPukenis (approved):

PR #52, opened 2023-05-29
Add storage FFI calls

Lipt0nas:


LukasPukenis (changes requested):
That's very nice. What is missing - Python wrappers for the new APIs and also at least one test case for each would be nice. Several for retrieval would be even nicer :)

norddrop/src/device/mod.rs
```diff
@@ -203,6 +203,71 @@ impl NordDropFFI {
...
+                .await
+                .as_mut()
+                .ok_or(ffi::types::NORDDROP_RES_NOT_STARTED)?
+                .purge_transfers_until(until_timestamp)
+                .await
+                .map_err(|_| ffi::types::NORDDROP_RES_DB_ERROR)
+        })
+    }
+
+    pub(super) fn get_transfers(&mut self, since_timestamp: i64) -> Result<String> {
```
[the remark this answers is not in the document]
matszczygiel: I would go even further `transfers_since` https://rust-lang.github.io/api-guidelines/naming.html#c-getter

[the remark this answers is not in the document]
Lipt0nas: fair enough, but I guess we should keep the binding-side naming with the `get` prefix

norddrop/src/device/mod.rs
```diff
@@ -203,6 +203,89 @@ impl NordDropFFI {
...
+                    }
+                })
+        })?;
+
+        let result =
+            serde_json::to_string(&result).map_err(|_| ffi::types::NORDDROP_RES_JSON_PARSE)?;
+
+        Ok(result)
+    }
+
```
matszczygiel, line 288: These are wrappers for wrappers. I propose keeping the `Storage` field in the `instance` and calling it directly. It will be less code and less mental overhead IMHO.

norddrop/src/device/mod.rs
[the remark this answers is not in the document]
matszczygiel: yeah, that's true ;)

norddrop/src/device/mod.rs
Lipt0nas: You mean adding a reference to `StorageDispatch` in `NorddropFFI` to skip going trough `Service` for the calls?

norddrop/src/ffi/mod.rs
```diff
@@ -260,6 +260,78 @@ pub extern "C" fn norddrop_stop(dev: &norddrop) -> norddrop_result {
...
+    result.unwrap_or(norddrop_result::NORDDROP_RES_ERROR)
+}
+
+/// Get all transfers since the given timestamp from the database. Accepts a
+/// UNIX timestamp in seconds
+#[no_mangle]
+pub extern "C" fn norddrop_get_transfers_since(
+    dev: &norddrop,
+    since_timestamp: i64,
+) -> *mut c_char {
```
matszczygiel, line 317: I'm unsure if returning `*mut c_char` is good here. With this in the `new_transfer` function, we already shoot ourselves in the foot. Because getting `null` can mean anything. 
Maybe we can do something like `transfers_since(dev: &norddrop, since: i64, out_transfers: **mut c_char) -> norddrop_result` and write the `char *` to the `out_transfers` on success? I'm not sure how swig will handle that TBH

LukasPukenis: I think this is a double-edged sword. The API would be better but I smell binding issues and problems by accepting pointer to pointer. This is fine to be left as tech debt for future revisions when we already have features going

matszczygiel: Yeah, it's probably true. But since it's potentially breaking change in the future I would at least try to see if Swig is a problem here.
If yes, then it's better to abandon the idea

Lipt0nas: I'll have a look if it's a trivial change or not :+1:

LukasPukenis: I think it's hard to quickly say if it's a good idea as Golang bindings have shown after some time that it causes trouble. Or C#. I suggest not blocking this PR due to this however exploring this idea for some small amount of time after it's been landed :)

Lipt0nas: I'll create a ticket for this as a possible improvement then, to keep it in the back of our minds

drop-transfer/src/service.rs
```diff
@@ -117,6 +117,33 @@ impl Service {
...
+        &mut self,
+        since_timestamp: i64,
+    ) -> Result<Vec<drop_storage::types::Transfer>, Error> {
+        self.state
+            .storage
+            .transfers_since(since_timestamp)
+            .await
+            .map_err(|_| Error::StorageError)
+    }
+
```
matszczygiel, line 146: Minor: to reduce wrapping you can just provide getter `pub fn storage(&self) -> &Storage` and delegate to it down the line

matszczygiel (approved):

drop-transfer/src/service.rs
Lipt0nas: Good catch on the `mut`, it remained from when it originally was an `Arc<Mutex<...>>`, As for the layers, i think id like to keep the ffi layer function clean, and leave the processing in `Service`

matszczygiel (approved):

PR #54, opened 2023-05-29
Remove outdated comment about NULL param to pubkey cb

LukasPukenis:
Public key callback initially passed in NULL
so that the host should return it's own pubkey, however that was changed and is no longer relevant.


Lipt0nas (approved):
Looks good

matszczygiel (approved):

matszczygiel:
@LukasPukenis let's merge it to unclutter the PRs page

PR #56, opened 2023-05-30
First unit test for drop-storage

LukasPukenis:


Lipt0nas (approved):
I misread the FAQ entry it seems, nice catch on the bug
And very nice that you updated the README too :+1:

PR #62, opened 2023-06-02
use `func (string) []byte` signature for go pubkey callback

matszczygiel:


LukasPukenis:
Can you elaborate how did you track down what's wrong?

PR #63, opened 2023-06-02
Lukasp/make ci green again

LukasPukenis:


drop-storage/src/lib.rs
```diff
@@ -24,9 +24,10 @@ pub struct Storage {
 impl Storage {
     pub async fn new(logger: Logger, path: &str) -> Result<Self> {
         let options = SqliteConnectOptions::from_str(path)?.create_if_missing(true);
-        let conn = SqlitePool::connect_with(options).await?;
+        let conn = SqlitePool::connect_with(options.clone()).await?;
+
         sqlx::migrate!("./migrations")
-            .run(&conn)
+            .run(&mut SqlitePool::connect_with(options).await?.acquire().await?)
```
matszczygiel, line 30: Why?

drop-storage/src/lib.rs
```diff
@@ -49,7 +50,7 @@ impl Storage {
             "INSERT OR IGNORE INTO peers (id) VALUES (?1)",
             transfer.peer
         )
-        .execute(&mut conn)
+        .execute(conn.acquire().await?)
```
matszczygiel, line 53: It should probably be `&mut *conn`

drop-storage/src/lib.rs
```diff
@@ -109,7 +116,7 @@ impl Storage {
             "INSERT INTO transfer_active_states (transfer_id) VALUES (?1)",
             transfer_id,
         )
-        .execute(&mut conn)
+        .execute(conn.acquire().await?)
```
matszczygiel, line 119: Same here. It's probably not necessary

drop-storage/src/lib.rs
LukasPukenis: It shouldn't compile if it would need a dereference?

LukasPukenis: Oh, I get it, it's probably that referencing allows me to omit `acquire()` part?

LukasPukenis: So I've tried it and yep `&mut *conn` works as well and looks a bit nicer so changed all occurences

drop-transfer/src/error.rs
```diff
@@ -116,6 +116,6 @@ impl From<walkdir::Error> for Error {
         value
             .into_io_error()
             .map(Into::into)
-            .unwrap_or(Error::BadPath("Filesystem loop detected".into()))
+            .unwrap_or_else(|| Error::BadPath("Filesystem loop detected".into()))
```
matszczygiel, line 119: :+1:

matszczygiel (approved):

PR #64, opened 2023-06-05
Review error handling in drop-transfer crate

matszczygiel:
I've introduced a description in the `BadPath`, `BadTransferState`, and `BadTransfer` errors and fixed their's usage.
I've also checked if the `Cancel` error is used correctly in the events and surprisingly there was only one place where it seemed incorrect


drop-transfer/src/error.rs
```diff
@@ -116,6 +116,6 @@ impl From<walkdir::Error> for Error {
         value
             .into_io_error()
             .map(Into::into)
-            .unwrap_or(Error::BadPath)
+            .unwrap_or(Error::BadPath("Filesystem loop detected".into()))
```
LukasPukenis, line 119: Loop?

LukasPukenis (approved):
oh yeah!

drop-transfer/src/error.rs
matszczygiel:
```rust
#[derive(Debug)]
enum ErrorInner {
    Io { path: Option<PathBuf>, err: io::Error },
    Loop { ancestor: PathBuf, child: PathBuf },
}
```
that's the `walkdir` error, so if it's not IO then it's loop (probably should never happen as we check for the links)

drop-transfer/src/ws/server/mod.rs
```diff
@@ -568,6 +568,10 @@ impl FileXferTask {
     ) {
         let init_res = match downloader.init(&self).await {
             Ok(init) => init,
+            Err(crate::Error::Canceled) => {
+                warn!(logger, "File cancelled on download init stage");
+                return;
+            }
```
Lipt0nas, line 574: So in this case if the file has already been cancelled, the application does not get an indication besides the log anymore?

drop-transfer/src/error.rs
LukasPukenis: Makes sense, thanks

drop-transfer/src/ws/server/mod.rs
LukasPukenis: That's a nice catch @Lipt0nas . I think the message is fine however the error mechanism should go back and be improved separately if at all

matszczygiel: I left a TODO comment, since this might become a non-issue when we implement the auto retries

PR #65, opened 2023-06-07
Store subpath in the DB

matszczygiel:
I'm hoping to include this one in 4.0 because it would be quite hard to restore somehow the subpath from the subsequent migration.
I've also included (in a second commit) making the transfer id strongly typed on the storage interface


drop-storage/src/types.rs
```diff
@@ -11,23 +11,42 @@ pub enum TransferType {
...
-    pub id: String,
-    pub path: String,
+pub struct TransferIncomingPath {
+    pub file_id: FileId,
+    pub relative_path: String,
     pub size: i64,
 }
 
+#[derive(Debug)]
+pub struct TransferOutgoignPath {
```
Lipt0nas, line 21:
```suggestion
pub struct TransferOutgoingPath {
```

matszczygiel: Done

Lipt0nas (approved):
Nice, I also like the change that moves the incoming/outgoing check to `storage_info`

drop-transfer/src/transfer.rs
```diff
@@ -88,18 +88,58 @@ impl Transfer {
     }
 
     pub fn storage_info(&self) -> StorageInfo {
+        // TODO(msz): this insane check wouldn't be needed if we had two different
+        // `Transfer` types
+        let is_incoming = match self.files.values().next() {
+            Some(files) => matches!(files.kind, FileKind::FileToRecv { .. }),
+            None => true, // TODO(msz): Arbitrarily chosen, there is no way to differentiate here
```
LukasPukenis, line 95: Not a very nice hack. Would be great to refactor the transfers. Mainly now there's some confusion as Transfers contain `files` that might be Incoming or Outgoing but then files themselves contain `kind` which is also `FileToSend` and `FileToRecv`. I will create a ticket for it

PR #67, opened 2023-06-07
Store download directory in database

Lipt0nas:
Also split up the started events into separate ones for downloading and uploading to make it a bit cleaner


drop-storage/migrations/20230406110818_initial_migration.sql
```diff
@@ -113,6 +113,7 @@ CREATE TABLE IF NOT EXISTS incoming_path_pending_states (
 CREATE TABLE IF NOT EXISTS incoming_path_started_states (
   id INTEGER PRIMARY KEY AUTOINCREMENT, 
   path_id INTEGER NOT NULL,
+  base_dir TEXT NOT NULL,
```
Lipt0nas, line 116: Not sure if this is the optimal naming, suggestions are welcome

matszczygiel: I would go for `dest_dir` but it's a soft suggestion as it's fine already

drop-storage/src/lib.rs
```diff
@@ -293,16 +293,18 @@ impl Storage {
         &self,
         transfer_id: Uuid,
         path_id: String,
+        base_dir: String,
```
matszczygiel, line 296: I'm not sure why we take the owned string in the DB methods, `&str` should suffice

drop-transfer/src/storage_dispatch.rs
```diff
@@ -40,22 +40,24 @@ impl<'a> StorageDispatch<'a> {
...
+
+            Event::FileDownloadStarted {
+                transfer_id,
+                file_id,
+                base_dir,
+            } => {
+                self.storage
+                    .insert_incoming_path_started_state(transfer_id, file_id, base_dir)
+                    .await?
+            }
```
matszczygiel, line 60: Nice :)

drop-transfer/src/storage_dispatch.rs
```diff
@@ -178,13 +180,14 @@ impl From<&crate::Event> for Event {
...
-                transfer_id: transfer.id(),
-                file_id: file.to_string(),
-            },
-            crate::Event::FileUploadStarted(transfer, file) => Event::Started {
-                transfer_type: TransferType::Outgoing,
+            crate::Event::FileDownloadStarted(transfer, file, base_dir) => {
+                Event::FileDownloadStarted {
+                    transfer_id: transfer.id(),
+                    file_id: file.to_string(),
+                    base_dir: base_dir.clone(),
```
matszczygiel, line 187: `
```suggestion
                    base_dir: base_dir,
```

drop-transfer/src/ws/server/mod.rs
```diff
@@ -412,11 +413,17 @@ async fn handle_client(
 }
 
 impl FileXferTask {
-    pub fn new(file: crate::File, xfer: crate::Transfer, location: PathBuf) -> crate::Result<Self> {
+    pub fn new(
+        file: crate::File,
+        xfer: crate::Transfer,
+        location: PathBuf,
+        base_dir: PathBuf,
```
matszczygiel, line 420: Isn't the location the same thing as base_dir ?

Lipt0nas: location here is the already glued together final path of the file

drop-transfer/src/storage_dispatch.rs
Lipt0nas: `base_dir` from the event is `&String`, can't not clone here

drop-storage/src/types.rs
```diff
@@ -49,10 +49,14 @@ pub enum Event {
     Pending {
         transfer_info: TransferInfo,
     },
-    Started {
-        transfer_type: TransferType,
+    FileUploadStarted {
+        transfer_id: TransferId,
+        file_id: FileId,
+    },
+    FileDownloadStarted {
```
LukasPukenis, line 56: :chef's kiss:

drop-transfer/src/ws/server/mod.rs
LukasPukenis: `full_location` or `absolute_path` are better names to describe `location`

LukasPukenis (dismissed):
There is some confusion around the naming of the fields that could be improved but that's not a blocker

drop-transfer/src/storage_dispatch.rs
matszczygiel: No, if it's `&String` then it's fine. I thought it was `String`

drop-transfer/src/ws/server/mod.rs
matszczygiel: Yes, the naming could be changed for a more descriptive one

matszczygiel: TBH we can get rid of `location` here as we can construct location as `base_dir` + `file.subpath()`

Lipt0nas: I agree with the naming, but getting rid of the location IMO might be a bit too of an intrusive change, as we don't trivially concatenate the paths, but also perform mapping and validation on it as well, so doing it once and storing, like we are doing now seems like a better solution

PR #70, opened 2023-06-12
Add path hash to JSON output

Lipt0nas:
Add more information to the JSON output for the app teams


matszczygiel:
TBH honest shouldn't that be called `file_id` ? I know that the name `path_hash` is used in multiple places but I wonder if it's something different than `file_id`

Lipt0nas:
> 

I mirrored the naming that is present on the database to avoid confusion on which fields we're actually retrieving, calling it `file_id` there might be a little confusing as the events have a FK 'path_id' that references the ID field of the path

LukasPukenis:
I would agree that `file_id` makes it more understandable. It doesn't need to be 1:1 mapping with the database but it also raises the question if the database field is named properly. `hash` implies the hash of the file not the hash of it's path. Maybe it's worth to rename it in additional MR as well?

LukasPukenis (dismissed):
ID usually leaves no questions that it's the mapping field, anything else incurs additional area for interpretation but it's really minor, it's good enough already. Please proceed as you wish :)

Lipt0nas:
If we don't intend to keep a 1:1 mapping here, then it's fine, I'll rename the field to `file_id` in the output

LukasPukenis (approved):

matszczygiel (approved):

PR #71, opened 2023-06-12
Clarify timestamp input values

Lipt0nas:
The Android team ran into an issue where passing `Long.MAX_VALUE` to `purge_transfers_until` would not delete any of the entries, and according to the SQLite documentation

`These functions only work for dates between 0000-01-01 00:00:00 and 9999-12-31 23:59:59 (julian day numbers 1721059.5 through 5373484.5). For dates outside that range, the results of these functions are undefined.`

To help avoid issues like this in the future, I have added the min and max bounds for the timestamps in the comment of the function


matszczygiel (dismissed):

Lipt0nas:
In addition, will add a check for these values in the code and return an error if it's out of bounds as well

PR #74, opened 2023-06-13
DB errors fix

matszczygiel:


drop-storage/src/lib.rs
```diff
@@ -707,24 +710,27 @@ impl Storage {
...
-                created_at: p.created_at.timestamp_millis(),
-                pending_states: vec![],
-                started_states: vec![],
-                cancel_states: vec![],
-                failed_states: vec![],
-                completed_states: vec![],
-            })
-            .collect::<Vec<_>>();
+        let mut paths = sqlx::query!(
+            r#"SELECT id as "path_id!", * FROM incoming_paths WHERE transfer_id = ?1"#,
```
LukasPukenis, line 714: Can you clarify the syntax a bit? It looks unfamiliar

matszczygiel:
Actually, it's not a simple query rewrite, it was wrong before because it made a subquery that could return 0 rows, and with `INSERT` it treated it as all NULL value single row. Whereas I rewrote it to use the `INSERT SELECT` query which works by inserting exactly the rows returned by `SELECT`.
It's a single statement now and not `INSERT` with subquery anymore

drop-storage/src/lib.rs
matszczygiel: It's the sqlx thing. It treated the id as `Option<>` which is not possible because it's `NOT NULL`. So I used the sqlx syntax to forcibly make it not an `Option<>`. It's a kind of `unwrap()` for sqlx

drop-storage/src/lib.rs
```diff
@@ -232,8 +233,10 @@ impl Storage {
         let mut conn = self.conn.acquire().await?;
 
         sqlx::query!(
-            "INSERT INTO outgoing_path_pending_states (path_id) VALUES ((SELECT id FROM \
```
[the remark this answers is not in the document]
matszczygiel: Actually, it's not a simple query rewrite, it was wrong before because it made a subquery that could return 0 rows, and with INSERT it treated it as all NULL value single row. Whereas I rewrote it to use the INSERT SELECT query which works by inserting exactly the rows returned by SELECT.
It's a single statement now and not INSERT with subquery anymore

drop-storage/src/lib.rs
LukasPukenis: How is `id` an Option when it's directly assigned to `IncomingPath.id` ? I don't really understand this id rename

drop-storage/src/lib.rs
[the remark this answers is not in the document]
matszczygiel: True, will add this too

drop-storage/src/lib.rs
matszczygiel: Exactly that's the problem that this is an `Option`. It's just sqlx decided for some reason that it's an Option, not me.
I mean I could've used `id as "id!"` here as well but the point is I needed to use the `!` to force sqlx into not making `Option` out of it

drop-storage/src/lib.rs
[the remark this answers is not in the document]
matszczygiel: Done

drop-storage/src/lib.rs
matszczygiel: It happens after I added the `UNIQE` constraint. Sqlx somehow cannot prove to itself that id cannot be NULL after the check is added

LukasPukenis (approved):

PR #75, opened 2023-06-13
Automatically remove database on error

matszczygiel:
I'm not sure if we want to do this on each runtime DB error

### DIFF
--- a/drop-core/src/status.rs
+++ b/drop-core/src/status.rs
@@ -1,0 +1,35 @@
+#[derive(Debug, Clone, Copy)]
+#[repr(u32)]
+pub enum Status {
+    Canceled = 1,
+    BadPath = 2,
+    BadFile = 3,
+    ServiceStop = 6,
+    BadTransfer = 7,
+    BadTransferState = 8,
+    BadFileId = 9,
+    Io = 15,
+    DirectoryNotExpected = 17,
+    TransferLimitsExceeded = 20,
+    MismatchedSize = 21,
+    UnexpectedData = 22,
+    InvalidArgument = 23,
+    TransferTimeout = 24,
+    WsServer = 25,
+    WsClient = 26,
+    AddrInUse = 27,
+    FileModified = 28,
+    FilenameTooLong = 29,
+    AuthenticationFailed = 30,
+    StorageError = 31,
+    DbLost = 32,
+}
+
+impl serde::Serialize for Status {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        (*self as u32).serialize(serializer)
+    }
+}
--- a/norddrop/src/device/mod.rs
+++ b/norddrop/src/device/mod.rs
@@ -128,14 +128,11 @@ impl NordDropFFI {
 
         let (tx, mut rx) = mpsc::channel::<drop_transfer::Event>(16);
 
-        let storage = Arc::new(self.rt.block_on(async {
-            drop_storage::Storage::new(self.logger.clone(), &self.config.drop.storage_path)
-                .await
-                .map_err(|e| {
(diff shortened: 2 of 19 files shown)


norddrop/src/device/mod.rs
```diff
@@ -558,3 +555,48 @@ fn prepare_transfer_files(
...
+            Ok(storage) => return Ok(storage),
+            Err(err) => {
+                i += 1;
+                warn!(logger, "Failed to open DB: {err}, already tried {i} times",);
+                if i >= 3 {
+                    break err;
+                }
+            }
+        }
+    };
```
LukasPukenis, line 577: Do we really suspect that opening the database might fail in a recoverable manner? This code looks fine so no harm to have it, just curious.

norddrop/src/device/types.rs
```diff
@@ -100,6 +100,9 @@ pub enum Event {
         #[serde(flatten)]
         data: FinishEvent,
     },
+    DbLost {
+        context: String,
+    },
```
LukasPukenis, line 105: Maybe it makes sense to wrap it into a generic `RuntimeError` type? This PR handles the database issues when opening it, but we will also encounter runtime errors with the database where we will need to recreate the database.

So we will have at least 2 database-related issues.

LukasPukenis (commented):
The code is nice in this PR and no hesitation to merge it, however, I want to highlight that we can fail at runtime as well with the database and should report an event in that case as well.

Could you do that alongside so that on any storage error, we do the removing + recreation part and also emit an error event?

And of course, an integration test per each case would be nice

norddrop/src/device/mod.rs
matszczygiel: Good question and the answer is I don't know. But removing the DB in case it is somehow recoverable would be a major offence I guess. So I just wanted to be safe

norddrop/src/device/types.rs
matszczygiel: Yeah, it makes sense. Let me do that

norddrop/src/device/mod.rs
LukasPukenis: That's a fair point, was just curious

matszczygiel:
@LukasPukenis take a look if it's more or less OK right now. I think that handling transfer-related DB errors should be left for another PR since we actually don't have a concise strategy for it

BTW I'm going to add one more commit containing the integration test

drop-core/src/status.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl serde::Serialize for Status {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        (*self as u32).serialize(serializer)
+    }
+}
```
LukasPukenis, line 35: Can't we decorate the enum with `#[derive(Serialize)]` ?

LukasPukenis (approved):
Looks neat! @Lipt0nas please take a look as well into this :) 

It makes perfect sense to handle transfer-deletion-from-db in a separate PR

drop-core/src/status.rs
matszczygiel: Um, yes. But it will be serialized as a string representation for example `DbLost` and not `32`.
TBH I would prefer the string variant but I was trying to be in line with the other events where the integer is used

LukasPukenis: Thanks for explaininig.

matszczygiel:
@LukasPukenis @Lipt0nas In general if we decide on proper error handling in all cases then I think the code responsible for that should eventually land in the `drop-storage` crate, for transparent error handling. It's also the place where we have the most information from the underlying sqlx errors.
But as I said it's the task for different PR and probably the next release

Lipt0nas (approved):
Looks good, another +1 from me

PR #77, opened 2023-06-14
Transfer must be inserted into the db at all times

matszczygiel:


drop-transfer/src/ws/client/mod.rs
```diff
@@ -223,8 +223,8 @@ impl RunContext<'_> {
         let (tx, rx) = mpsc::unbounded_channel();
 
         let mut lock = self.state.transfer_manager.lock().await;
-        lock.insert_transfer(self.xfer.clone(), TransferConnection::Client(tx))
-            .await?;
+        lock.insert_transfer(self.xfer.clone(), TransferConnection::Client(tx))?;
+        drop(lock);
```
Lipt0nas, line 227: Is there a reason this can't be a one-liner?

LukasPukenis (commented):
It looks like now the TransferManager and Storage receive the transfer at different times. TransferManager inserting to storage is less error prone, can it be tweaked this way?

matszczygiel:
@LukasPukenis the close coupling was exactly the reason the issue was there in the first place

drop-transfer/src/ws/client/mod.rs
matszczygiel: Probaably

matszczygiel: done

LukasPukenis:
@matszczygiel what I mean is that manager knows about the transfer earlier than the storage or vice-versa so now it might manifest in another race condition down the line as they're tightly related

drop-transfer/src/ws/client/mod.rs
Lipt0nas: I think you forgot to add the changes before pushing :sweat_smile:

LukasPukenis (approved):
Looks solid

PR #79, opened 2023-06-15
Insert transfer paths in case the app is using FDs

matszczygiel:


drop-transfer/src/transfer.rs
```diff
@@ -111,25 +111,27 @@ impl Transfer {
...
+                        FileKind::FileToSend { source, .. } => match source {
+                            FileSource::Path(fullpath) => fullpath
+                                .ancestors()
+                                .nth(f.subpath.iter().count())
+                                .and_then(|rp| rp.to_str())
+                                .unwrap_or("."),
+                            FileSource::Fd(_) => ".", /* Let's pretend the files are in the
+                                                       * working dir. The FDs are only used on
+                                                       * Android, */
+                        },
```
LukasPukenis, line 125: What implications using `.` does it have? Asking about both cases

matszczygiel: Yeah, the implications are such that the resume won't work with this kind of path. The DB stores the `base_path` which is not available for Android so I had to store "something".

LukasPukenis: is there a case where the first one can fail as it contains `unwrap_or`? It's going to be executed in non-android cases and the previous code didn't have such a guard.

As for Android - I am on the verge of suggesting making the path nullable to encode this case instead. What do you think?

LukasPukenis: Thanks :)

For non-android case `filter-map` looks like a nice solution given that we have plenty of logging in erroneous cases as we would like to track it down in case paths will not be stored. That needs to be addressed.

For Android case to handle it I think the nullable column in sqlite is the way to model it.

matszczygiel: Yeah, I agree on filter_map but do we really need `base_path` nullable?

matszczygiel: I restored the filter_map but I'm still hesitant about the nullable column

LukasPukenis: I don't really want a nullable column but... we need to encode Android somehow. The current solution is to write a `.` instead which we need to be aware off. Do you think it's a big effort to nullify the column? If so the dot is fine, it's just the Android path of logic gets more and more complex through time :/

PR #80, opened 2023-06-15
Merge database JSON events into one list

Lipt0nas:
Even more spring cleaning for the JSON output

- Reordered some of the fields to make the output a bit more efficient to visually parse
- Merge both the path and transfer events into one list and sort them by timestamp

And as a side effect-the `types.rs` for storage is a little less bloated

### DIFF
--- a/drop-storage/src/lib.rs
+++ b/drop-storage/src/lib.rs
@@ -6,8 +6,9 @@ use std::str::FromStr;
 use slog::Logger;
 use sqlx::{sqlite::SqliteConnectOptions, SqliteConnection, SqlitePool};
 use types::{
-    DbTransferType, IncomingPath, OutgoingPath, PathStateEvent, PathStateEventData, Transfer,
-    TransferFiles, TransferIncomingPath, TransferOutgoingPath, TransferStateEvent,
+    DbTransferType, IncomingPath, IncomingPathStateEvent, IncomingPathStateEventData, OutgoingPath,
+    OutgoingPathStateEvent, OutgoingPathStateEventData, Transfer, TransferFiles,
+    TransferIncomingPath, TransferOutgoingPath, TransferStateEvent,
 };
 use uuid::{fmt::Hyphenated, Uuid};
 
@@ -168,8 +169,7 @@ impl Storage {
             tid,
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
@@ -185,8 +185,7 @@ impl Storage {
             error,
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
@@ -206,8 +205,7 @@ impl Storage {
             by_peer,
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
@@ -230,8 +228,7 @@ impl Storage {
             file_id
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
@@ -252,8 +249,7 @@ impl Storage {
             file_id,
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
@@ -275,8 +271,7 @@ impl Storage {
             0
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
@@ -300,8 +295,7 @@ impl Storage {
             0
         )
         .execute(&mut *conn)
-        .await
-        .map_err(error::Error::DBError)?;
+        .await?;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


drop-storage/src/lib.rs
```diff
@@ -552,49 +546,62 @@ impl Storage {
...
+                    created_at: s.created_at.timestamp_millis(),
+                    data: types::TransferStateEventData::Failed {
+                        status_code: s.status_code,
+                    },
+                }),
+            );
+
+            transfer
+                .states
+                .sort_by(|a, b| a.created_at.cmp(&b.created_at));
```
matszczygiel, line 604: The sqlx gives you the `NaiveDateTime` which has a better precision than `timestamp_millis()` so maybe we should sort by the original thing and then convert it to ms or just use the conversions during the serialization with `#[serde(serialize_with = "NaiveDateTime::timestamp_millis()")]`

drop-storage/src/lib.rs
```diff
@@ -552,49 +546,62 @@ impl Storage {
...
-            })
-            .collect();
+            transfer.states.extend(
+                sqlx::query!(
+                    "SELECT created_at  FROM transfer_active_states WHERE transfer_id = ?1",
+                    tid
+                )
+                .fetch_all(&mut *conn)
+                .await
+                .map_err(error::Error::DBError)?
```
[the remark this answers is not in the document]
matszczygiel: There is a lot of these

drop-storage/src/types.rs
```diff
@@ -3,6 +3,66 @@ use serde::Serialize;
 type TransferId = uuid::Uuid;
 type FileId = String;
 
+#[derive(Debug, Serialize)]
+#[serde(tag = "state")]
+pub enum PathStateEventData {
```
matszczygiel, line 8: This should be split into two types, one for incoming and the other for outgoing IMHO

matszczygiel:
I like the JSON structure change, but I think it should be achieved somewhat differently in the code - I wrote some comments

drop-storage/src/lib.rs
matszczygiel: If that's the case then there is no point in keeping the `NaiveDateTime`. But still, I think that using the sqlx native type would be beneficial because then we can then use the `sqlx::query_as!()` macro with the `types.rs` structures directly.
I'm leaving the suggestion for you to decide

drop-storage/src/types.rs
Lipt0nas: done :+1:

drop-storage/src/lib.rs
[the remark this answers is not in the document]
Lipt0nas: changed

drop-storage/src/lib.rs
Lipt0nas: I changed the fields to be `NaiveDateTime`, though it seems to use `query_as!` there would need to be quite a big change to the queries, ill come back to that in a different MR in the future, since it's not a pressing concern right now

drop-storage/src/lib.rs
```diff
@@ -726,92 +724,99 @@ impl Storage {
...
+                .await?
+                .iter()
+                .map(|s| IncomingPathStateEvent {
+                    path_id: s.path_id,
+                    created_at: s.created_at,
+                    data: IncomingPathStateEventData::Started {
+                        base_dir: s.base_dir.clone(),
+                        bytes_received: s.bytes_received,
+                    },
+                }),
```
matszczygiel, line 763:
```suggestion
                sqlx::query!(
                    "SELECT * FROM incoming_path_started_states WHERE path_id = ?1",
                    path.id
                )
                .fetch_all(&mut *conn)
                .await?
                .into_iter()
                .map(|s| IncomingPathStateEvent {
                    path_id: s.path_id,
                    created_at: s.created_at,
                    data: IncomingPathStateEventData::Started {
                        base_dir: s.base_dir,
                        bytes_received: s.bytes_received,
                    },
                }),
```

drop-storage/src/lib.rs
```diff
@@ -726,92 +724,99 @@ impl Storage {
...
+                .fetch_all(&mut *conn)
+                .await?
+                .iter()
+                .map(|s| IncomingPathStateEvent {
+                    path_id: s.path_id,
+                    created_at: s.created_at,
+                    data: IncomingPathStateEventData::Completed {
+                        final_path: s.final_path.clone(),
+                    },
+                }),
```
matszczygiel, line 816:
```suggestion
                .into_iter()
                .map(|s| IncomingPathStateEvent {
                    path_id: s.path_id,
                    created_at: s.created_at,
                    data: IncomingPathStateEventData::Completed {
                        final_path: s.final_path,
                    },
                }),
```

LukasPukenis (commented):
+1 it looks nice, except those `map(|_| ())` things

drop-storage/src/lib.rs
```diff
@@ -102,9 +99,9 @@ impl Storage {
             path.size,
         )
         .execute(conn)
-        .await?;
-
-        Ok(())
+        .await
+        .map_err(error::Error::DBError)
+        .map(|_| ())
```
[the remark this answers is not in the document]
Lipt0nas: reverted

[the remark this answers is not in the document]
matszczygiel: Did you push the changes? It doesn't look like you did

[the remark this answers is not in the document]
Lipt0nas: I managed to forget to save the file :see_no_evil:, my bad

matszczygiel:
No redundant `clone()` on my radar. +1

matszczygiel (approved):

PR #81, opened 2023-06-15
Lukasp/storage layer logging

LukasPukenis:
This PR introduces storage layer logging as well as query logging from `sqlx`. For this to work `tracing` is marshalled through a global logger and then marshalled to `slog` so we would be able to drain those logs via callback

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ members = [
 
 [workspace.dependencies]
 tokio = { version = "1", features = ["rt-multi-thread", "net", "sync", "macros"] }
-slog = { version = "2.7.0", features = ["release_max_level_info"] }
+slog = { version = "2.7.0", features = ["release_max_level_info", "max_level_trace"] }
 serde = { version = "1.0.160", features = ["derive"] }
 async-trait = "0.1.68"
 base64 = "0.21.0"
--- a/drop-storage/src/lib.rs
+++ b/drop-storage/src/lib.rs
@@ -8,7 +8,7 @@ use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 use rusqlite::{params, Transaction};
 use rusqlite_migration::Migrations;
-use slog::Logger;
+use slog::{trace, Logger};
 use types::{
     DbTransferType, IncomingPath, IncomingPathStateEvent, IncomingPathStateEventData, OutgoingPath,
     OutgoingPathStateEvent, OutgoingPathStateEventData, Transfer, TransferFiles,
@@ -23,8 +23,8 @@ type Result<T> = std::result::Result<T, Error>;
 type QueryResult<T> = std::result::Result<T, rusqlite::Error>;
 // SQLite storage wrapper
 pub struct Storage {
-    _logger: Logger,
     pool: Pool<SqliteConnectionManager>,
+    logger: Logger,
 }
 
 const MIGRATIONS_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/migrations");
@@ -45,10 +45,7 @@ impl Storage {
             .to_latest(&mut conn)
             .map_err(|e| Error::InternalError(format!("Failed to run migrations: {e}")))?;
 
-        Ok(Self {
-            _logger: logger,
-            pool,
-        })
+        Ok(Self { logger, pool })
     }
 
     pub fn insert_transfer(&self, transfer: &TransferInfo) -> Result<()> {
@@ -58,6 +55,12 @@ impl Storage {
         };
 
         let tid = transfer.id.to_string();
+        trace!(
+            self.logger,
+            "Inserting transfer";
+            "transfer_id" => &tid,
+            "transfer_type" => transfer_type_int,
+        );
 
         let mut conn = self.pool.get()?;
         let conn = conn.transaction()?;
@@ -69,11 +72,23 @@ impl Storage {
 
         match &transfer.files {
             TransferFiles::Incoming(files) => {
+                trace!(
+                    self.logger,
+                    "Inserting transfer::Incoming files len {}",
+                    files.len()
+                );
+
                 for file in files {
                     Self::insert_incoming_path(&conn, transfer.id, file)?;
                 }
             }
             TransferFiles::Outgoing(files) => {
+                trace!(
+                    self.logger,
+                    "Inserting transfer::Outgoing files len {}",
+                    files.len()
+                );
+
                 for file in files {
                     Self::insert_outgoing_path(&conn, transfer.id, file)?;
                 }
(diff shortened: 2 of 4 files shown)


drop-storage/src/lib.rs
```diff
@@ -47,6 +51,13 @@ impl Storage {
             TransferFiles::Outgoing(_) => TransferType::Outgoing as u32,
         };
 
+        info!(
+            self._logger,
```
matszczygiel, line 55: We should rename `_logger` -> `logger` I guess

drop-storage/src/lib.rs
```diff
@@ -69,12 +80,33 @@ impl Storage {
 
         match &transfer.files {
             TransferFiles::Incoming(files) => {
+                info!(self._logger, "incoming files len {}", files.len());
+
                 for file in files {
+                    info!(
```
matszczygiel, line 86: I would say let's change it to `debug!()` because we're logging here for example paths, which was not good from the security point of view

norddrop/src/ffi/mod.rs
```diff
@@ -373,6 +373,41 @@ pub extern "C" fn norddrop_new(
     });
 
     let result = panic::catch_unwind(move || {
+        {
```
matszczygiel, line 376: I would say, let's use it under the feature flag. We pretty much don't want this in the release builds

norddrop/src/ffi/mod.rs
```diff
@@ -373,6 +373,41 @@ pub extern "C" fn norddrop_new(
...
+                fn log(&self, record: &Record) {
+                    if self.enabled(record.metadata()) {
+                        match record.level() {
+                            Level::Error => slog::error!(self.logger, "{}", record.args()),
+                            Level::Warn => slog::warn!(self.logger, "{}", record.args()),
+                            Level::Info => slog::info!(self.logger, "{}", record.args()),
+                            Level::Debug => slog::debug!(self.logger, "{}", record.args()),
+                            Level::Trace => slog::trace!(self.logger, "{}", record.args()),
+                        }
+                    }
```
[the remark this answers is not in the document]
matszczygiel: Or on the other hand, we can do the opposite and use the log throughout the library and get rid of `slog`. It might be even easier in the long run

drop-storage/src/lib.rs
LukasPukenis: debug! makes sense here

norddrop/src/ffi/mod.rs
LukasPukenis: It will invoke `slog` macros which are going to be compiled out. `slqx` will also use the provided severity so it's pretty good. I've added additional guard just in case

norddrop/src/ffi/mod.rs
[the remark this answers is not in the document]
matszczygiel: Yeah, I meant using exactly the `logger.log()` method. It's a method from the `Drain` trait

Cargo.toml
```diff
@@ -13,7 +13,7 @@ members = [
 
 [workspace.dependencies]
 tokio = { version = "1", features = ["rt-multi-thread", "net", "sync", "macros"] }
-slog = { version = "2.7.0", features = ["release_max_level_info"] }
+slog = { version = "2.7.0", features = ["release_max_level_info", "std"] }
```
matszczygiel, line 16: `std` feature is default. No need to explicitly enable it

drop-storage/src/lib.rs
```diff
@@ -18,24 +18,29 @@ pub use crate::types::{Event, FileChecksum, TransferInfo, TransferType};
...
+        #[allow(unused_mut)]
+        let mut options = SqliteConnectOptions::from_str(path)?.create_if_missing(true);
+
+        #[cfg(debug_assertions)]
+        {
+            use sqlx::ConnectOptions;
+            options = options.log_statements(log::LevelFilter::Debug);
+        }
+
+        let conn = SqlitePool::connect_with(options.clone()).await?;
```
Lipt0nas, line 36:
```suggestion
        let conn = SqlitePool::connect_with(options).await?;
```

matszczygiel, line 36:
```suggestion
        let conn = SqlitePool::connect_with(options).await?;
```

drop-storage/src/lib.rs
```diff
@@ -177,6 +225,12 @@ impl Storage {
     pub async fn insert_transfer_failed_state(&self, transfer_id: Uuid, error: u32) -> Result<()> {
         let tid = transfer_id.hyphenated();
 
+        info!(
+            self.logger,
+            "Inserting transfer failed state";
+            "transfer_id" => tid.to_string(),
```
matszczygiel, line 231: Do we actually implement structured logging with the log callback? I guess not

drop-storage/src/lib.rs
matszczygiel: Nice timing :+1:

drop-storage/Cargo.toml
```diff
@@ -12,5 +12,6 @@ slog = { workspace = true }
 serde = { workspace = true }
 uuid = { workspace = true }
 
+log = "0.4.19"
```
Lipt0nas, line 15:
```suggestion
log = "0.4.19"

```

matszczygiel:
TBH, do query logging solves any problem for us? I feel like it's best to get rid of log and tracing and keep our `debug!(self.logger, ...)` macros. But given that the code is already written I would say put it under the feature flag in case it will be needed in the future

LukasPukenis:
I've left it simple - trace logs in the storage crate to track down what is happening and that's it. I reasoned about `rows_affected` however questioned their value as well in the end and thus didn't add those for now

matszczygiel:
@LukasPukenis BTW slog disables trace logs in the debug build https://docs.rs/slog/latest/slog/#notable-details maybe we can turn them on?

LukasPukenis:
@matszczygiel Actually I was wondering why trace logs don't appear any more and just left it there in favor of a release. Thanks for the link

drop-storage/src/lib.rs
LukasPukenis: That was a  good catch, my eyes missed that one. It is now implemented in our drain

LukasPukenis:
@matszczygiel as per your recommendation - I've turned on tracing logs and implemented k:v logging as well. Sprinkled trace logs all over storage crate as well. I think those logs borderline on being `debug!` but on release builds I think they would just trash the logs too much

matszczygiel:
Everything looks good :+1:  but to be honest when I run `scenario1` I don't see the storage logs. Am I doing something wrong?

LukasPukenis:
@matszczygiel that's becasue `ffi.py` still has `debug` severity, however, I think for the time being we can enable trace logs in there as well :)

matszczygiel:
Oh, I haven't noticed. Great job :+1:

matszczygiel (approved):

PR #82, opened 2023-06-15
Remove peers table

matszczygiel:
This table does literally nothing. Let's not clutter the DB


LukasPukenis (approved):
Please don't forget to update the db schema in docs as well

PR #83, opened 2023-06-15
Resumes for paths with directories

matszczygiel:


drop-transfer/src/manager.rs
```diff
@@ -25,6 +28,7 @@ pub struct TransferState {
     pub(crate) connection: TransferConnection,
     // Used for mapping directories inside the destination
     dir_mappings: HashMap<PathBuf, String>,
+    dir_maker_file_name: String,
```
Lipt0nas, line 31:
```suggestion
    dir_marker_file_name: String,
```

Lipt0nas (approved):
@LukasPukenis should also take a look, I really like the gradual reduction of `moose_try_file` calls as well

matszczygiel:
Let's wait with merging this one for https://github.com/NordSecurity/libdrop/pull/72 since it might introduce some nuances into the directory resumes

test/scenarios.py
```diff
@@ -4090,6 +4090,188 @@
...
+                        )
+                    ),
+                    action.WaitRacy(
+                        [
+                            event.Start(0, FILES["nested/big/testfile-01"].id),
+                            event.Start(0, FILES["nested/big/testfile-02"].id),
+                        ]
+                    ),
+                    action.Wait(event.FinishTransferCanceled(0, True)),
+                    action.WaitForAnotherPeer(),
```
LukasPukenis, line 4126: `WaitForAnotherPeer` is intended to be at the top of the scenario to ensure the other side has booted up, in here it servers just as an unnecessary delay it seems.

matszczygiel: Yes, I probably pasted this here by accident. Thanks :+1:

matszczygiel: Done

test/scenarios.py
```diff
@@ -4057,6 +4057,187 @@
             ),
         },
     ),
+    Scenario(
```
LukasPukenis, line 4060: Could you add a test that sends `/tmp/nested`  so to test the multiple levels of directories?

test/scenarios.py
```diff
@@ -4057,6 +4057,187 @@
...
+                                event.File(
+                                    FILES["nested/big/testfile-02"].id,
+                                    "big/testfile-02",
+                                    10485760,
+                                ),
+                            },
+                        )
+                    ),
+                    action.WaitRacy(
+                        [
```
LukasPukenis, line 4113: But this actually sends two separate requests and tests that those two requests were sent properly. We don't even test that there should be progress which is equal/higher to the one that was previously reported. The IDs also will differ so the marker file will be different so it shouldn't really resume?

drop-transfer/src/manager.rs
```diff
@@ -103,20 +78,36 @@ impl TransferManager {
...
-                        vacc.insert(
+                        let mapped = crate::utils::filepath_variants(vacc.key())?.find(|dst_location| {
+                                // Skip if there is already a file with the same name.
+                                // Additionaly there could be a dangling symlink with the same name,
+                                // the `symlink_metadata()` ensures we can catch that.
+                                matches!(dst_location.symlink_metadata() , Err(err) if err.kind() == io::ErrorKind::NotFound) ||
+                                // Or it might be our dir created by the same transfer (as this can be a resume)
+                                dst_location.join(&self.dir_maker_file_name).symlink_metadata().is_ok()
+                            })
+                            .expect("The filepath variants iterator should never end");
```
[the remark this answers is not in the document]
matszczygiel: Yes. Let me try

drop-transfer/src/manager.rs
```diff
@@ -132,28 +123,69 @@ impl TransferManager {
...
+
+    pub(crate) fn state(&self, id: Uuid) -> Option<&TransferState> {
+        self.transfers.get(&id)
+    }
+
+    pub(crate) fn state_mut(&mut self, id: Uuid) -> Option<&mut TransferState> {
+        self.transfers.get_mut(&id)
+    }
+}
+
```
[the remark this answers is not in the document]
matszczygiel: Convenience basically. The `TransferManager` started to become cluttered with a lot of checks if a transfer exists for each call. And it was frequent that the functions were called in a series for a single transfer so it:
* reduces the number of redundant checks and potential error paths
* simplifies the interface

It's a minor refactor

drop-transfer/src/manager.rs
```diff
@@ -132,28 +123,69 @@ impl TransferManager {
...
 pub(crate) struct TransferGuard {
-    state: ManuallyDrop<Arc<State>>,
+    state: Option<Arc<State>>,
```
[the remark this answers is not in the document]
matszczygiel: It was actually needed because of the graceful shutdown which needs to disable somehow running the "drop" code

test/scenarios.py
matszczygiel: Yeah, so there is one more thing. I moved the dir creation code later during the transfer. It was done upload calling download. But I moved it when the downloaded temporary is placed into the destination.
So basically this test case tests nothing related to the directory resume TBH.
I must have missed it somehow

drop-transfer/src/manager.rs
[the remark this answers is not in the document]
matszczygiel: Done

drop-storage/src/lib.rs
```diff
@@ -691,6 +692,26 @@ impl Storage {
         Ok(transfers)
     }
 
+    pub async fn finished_files(&self, transfer_id: Uuid) -> Result<Vec<FinishedFile>> {
```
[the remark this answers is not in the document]
matszczygiel: Done

matszczygiel:
I removed the marker files created from this PR (in a separate commit so it can be reverted later in any case)

test/scenarios.py
matszczygiel: I've changed the test case to use `/tmp/nested`  dir

LukasPukenis (approved):

PR #84, opened 2023-06-16
Include transfer uuid in the V4 temporary file for Android

matszczygiel:


LukasPukenis (approved):
Not gonna lie, this looks dope

PR #85, opened 2023-06-16
Update docs

LukasPukenis:
Our documentation is generated straight from Rust code however we have leftovers in `norddrop.i`. This is cleaned up to reduce confusion in the future.

JSON descriptions, examples and all parameters
are commented on FFI calls allowing us to
just copy-paste the generated `norddrop.h`
straight into the documentation.

### DIFF
--- a/norddrop/src/ffi/mod.rs
+++ b/norddrop/src/ffi/mod.rs
@@ -39,6 +39,31 @@ extern "C" {
 #[allow(non_camel_case_types)]
 pub struct norddrop(Mutex<NordDropFFI>);
 
+/// @brief Initialize a new transfer with the peer and descriptors provided and
+/// return the transfer ID
+///
+/// @param dev   Pointer to the instance
+/// @param peer  Peer address
+/// @param descriptors   JSON descriptors
+/// @return char*  Transfer ID
+///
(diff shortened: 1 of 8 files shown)


LukasPukenis:
SWIG doesnt understand `/**/` comments, so `///` is used in ffi layer

norddrop/src/ffi/mod.rs
```diff
@@ -39,6 +39,31 @@ extern "C" {
 #[allow(non_camel_case_types)]
 pub struct norddrop(Mutex<NordDropFFI>);
 
+/// @brief Initialize a new transfer with the peer and descriptors provided and
```
matszczygiel, line 42: What kind of syntax is this? Are you aware that rust uses markdown for doc comments?

LukasPukenis: doxygen

matszczygiel (approved):

norddrop/src/ffi/mod.rs
LukasPukenis: Reading https://doc.rust-lang.org/rust-by-example/meta/doc.html now it makes sense why it only supports ///.

PR #89, opened 2023-06-19
Make rejection out of cancel

matszczygiel:
This PR basically changes the cancel behavior:
* events about the cancellation are emitted always
* cancel is done once and for good

### DIFF
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
     mem::ManuallyDrop,
     path::{Path, PathBuf},
     sync::Arc,
@@ -25,6 +25,8 @@ pub struct TransferState {
     pub(crate) connection: TransferConnection,
     // Used for mapping directories inside the destination
     dir_mappings: HashMap<PathBuf, String>,
+
+    cancelled: HashSet<FileId>,
 }
 
 /// Transfer manager is responsible for keeping track of all ongoing or pending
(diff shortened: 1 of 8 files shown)


drop-transfer/src/manager.rs
```diff
@@ -76,6 +79,27 @@ impl TransferManager {
...
+            .transfers
+            .get_mut(&id)
+            .ok_or(crate::Error::BadTransfer)?;
+        if !xstate.xfer.files().contains_key(&file) {
+            return Err(crate::Error::BadFileId);
+        }
+
+        if !xstate.cancelled.insert(file) {
+            return Err(crate::Error::FileCancelled);
+        }
```
Lipt0nas, line 98: Since double cancelling doesn't functionally break anything here, and the app teams rely on events always being emitted makes me think, should we return an error here?

matszczygiel: Good point. IDK. @LukasPukenis what do you think?

LukasPukenis: Both ways seem to be valid approaches :)

We might leave it as it was but provide an additional flag to indicate if the file was in transit when the cancellation happened. This way:
- we always emit the event
- indicate if the file was in flight, so no information hiding is present


I am personally leaning toward what I've described above as we don't break the API needlessly and it sounds simpler. We can also release it as v4.* instead of waiting for v5

@matszczygiel As there's no clear winner in this case I suggest proceeding as you wish

matszczygiel: @Lipt0nas @LukasPukenis I'm not sure if we all have the same picture.
My understanding is this: the cancel is no longer related to the file being in transit (well it can potentially stop the transit if the file is in flight). Cancel is meant for disabling file transit and it cannot be undone.
So the question is: Since the cancel is idempotent, should we emit an error if the cancel is called more than once?

matszczygiel:
Closing in favor of https://github.com/NordSecurity/libdrop/pull/90

PR #90, opened 2023-06-20
Introduce rejections

matszczygiel:
Introducing V5 wire protocol with rejection support. On older protocols, there is a fallback to cancelation.
Currently what I did for older protocols is that cancel is issued and if peer requests download after that it will receive the error message.
It can be changed so that we can emit an error upon calling `reject()` if our peer does not support rejection, what do you think?

The old <-> new protocol is actually not tested because TBH there is not an easy way to issue rejection from udrop. It would require proper manual testing on actual integration or to somehow devise a test framework for testing wire compatibility which we don't have at this moment.

There is one thing still missing. The temporaries are not deleted after rejection and in general, after the transfer is failed or terminated before all files are downloaded. It's a wider problem and deserves a PR on its own IMHO

### DIFF
--- a/drop-storage/src/types.rs
+++ b/drop-storage/src/types.rs
@@ -174,12 +174,12 @@ pub enum Event {
         file_id: FileId,
         final_path: String,
     },
-    Progress {
+    FileProgress {
         transfer_id: TransferId,
         file_id: FileId,
         progress: i64,
     },
-    Reject {
+    FileReject {
         transfer_type: TransferType,
         transfer_id: TransferId,
         file_id: FileId,
--- a/drop-storage/src/types.rs
+++ b/drop-storage/src/types.rs
@@ -180,6 +180,7 @@ pub enum Event {
         progress: i64,
     },
     Reject {
+        transfer_type: TransferType,
         transfer_id: TransferId,
         file_id: FileId,
         by_peer: bool,
--- a/drop-storage/src/types.rs
+++ b/drop-storage/src/types.rs
@@ -179,6 +179,11 @@ pub enum Event {
         file_id: FileId,
         progress: i64,
     },
+    Reject {
+        transfer_id: TransferId,
+        file_id: FileId,
+        by_peer: bool,
+    },
 }
 
 #[derive(Debug, Serialize)]
--- a/drop-transfer/src/ws/client/v2.rs
+++ b/drop-transfer/src/ws/client/v2.rs
@@ -96,6 +96,60 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
         Ok(())
     }
 
+    async fn issue_reject(
+        &mut self,
+        socket: &mut WebSocket,
+        file_id: FileId,
+    ) -> anyhow::Result<()> {
+        let file_subpath = if let Some(file) = self.xfer.files().get(&file_id) {
+            file.subpath().clone()
+        } else {
+            warn!(self.logger, "Missing file with ID: {file_id:?}");
+            return Ok(());
+        };
+
+        let msg = v2::ClientMsg::Cancel(v2::Download {
+            file: file_subpath.clone(),
+        });
+        socket.send(Message::from(&msg)).await?;
+
+        if let Some(task) = self.tasks.remove(&file_subpath) {
+            if !task.job.is_finished() {
+                task.job.abort();
+
+                task.events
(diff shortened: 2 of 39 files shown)


drop-storage/src/types.rs
```diff
@@ -179,6 +179,12 @@ pub enum Event {
         file_id: FileId,
         progress: i64,
     },
+    Reject {
+        transfer_type: TransferType,
+        transfer_id: TransferId,
+        file_id: FileId,
+        by_peer: bool,
+    },
```
LukasPukenis, line 187: The naming is inconsistent. The same goes for `Progress`. I think adding `File` prefix would be nice here

drop-transfer/src/ws/client/v2.rs
```diff
@@ -96,6 +96,60 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
...
+            if !task.job.is_finished() {
+                task.job.abort();
+
+                task.events
+                    .stop(crate::Event::FileUploadCancelled(
+                        self.xfer.clone(),
+                        file_id.clone(),
+                        false,
+                    ))
+                    .await;
```
LukasPukenis, line 126: It's confusing to see which version works where. This is `v2` file so I would assume this is v2 logic in here however it's clearly v5 logic inside of this file.

Another point - rejection fails if the cancel fails. I think rejection should follow through fully locally despite any errors on the other end. This is only for this file as client/v4 and client/v5 have it without cancel.

it looks like this should be `on_reject` instead of `issue_reject`. TLDR: architecture is very confusing to an external reader, could you clarify things?

drop-transfer/src/ws/client/v4.rs
```diff
@@ -156,7 +203,15 @@ impl HandlerLoop<'_> {
         file_id: FileId,
         limit: u64,
     ) -> anyhow::Result<()> {
-        let f = || {
+        let f = async {
```
LukasPukenis, line 206: Could you explain this change?

LukasPukenis (commented):
Rejection logic is for v5 only, I am not sure why v2 and v4 were updated at all. Overall rejection logic makes sense but there's a confusion in architecture. I've left a comment asking you to explain it

drop-transfer/src/ws/client/v4.rs
matszczygiel: I needed to use `.await` inside of lambda and that's the way to go

drop-transfer/src/ws/client/v2.rs
matszczygiel: Actually, V4 does not have rejections and it's the same logic.
I wrote in the opening comment that there are two approaches:
* fallback to cancel
* Return to the caller with the error

I picked the first one. Can switch to the second option quite painlessly

`on_x()` method generally contains the handler code for something that happened, e.g. when there is a download request you have the `on_download()` handler
`issue_x()` is when the caller requests something to happen, so, in this case, it is correctly named as `issue_reject()` since you cannot receive from peer rejection because the wire protocol does not support that

drop-storage/src/types.rs
matszczygiel: Done

drop-transfer/src/ws/client/v2.rs
LukasPukenis: I think I got it. `client/vN.rs` are client code when communicating with that version, for example, `client/v2.rs` might still be used if we communicate with v2 server? In that case it would make sense

matszczygiel: Yes, `client/v2.rs` is a client implementation for communicating with the V2 protocol. The same goes for the server.

So I presented in this PR the fallback for cancellation in the case of V2 and V4 protocols.
Do you think that's OK? We can switch to reporting errors to the caller, effectively disabling this functionality for communicating with legacy peers

matszczygiel: The conclusion is that for this moment we fallback to cancel

LukasPukenis (dismissed):
I've looked through test cases and API, both make sense. The tests are very nice. I however hope for the best on the protocol level as there are a lot of changes but at a glance, they look good

@Lipt0nas should take a look as well

matszczygiel:
The PR was approved before minor changes, so I'm merging this one

PR #91, opened 2023-06-21
Add checksum validation at the end of download

matszczygiel:


drop-core/src/status.rs
```diff
@@ -23,6 +23,7 @@ pub enum Status {
     AuthenticationFailed = 30,
     StorageError = 31,
     DbLost = 32,
+    FileChecsumMismatch = 33,
```
Lipt0nas, line 26:
```suggestion
    FileChecksumMismatch = 33,
```

drop-transfer/src/error.rs
```diff
@@ -46,6 +46,8 @@ pub enum Error {
     AuthenticationFailed,
     #[error("Storage error")]
     StorageError,
+    #[error("Checksum validation failed")]
+    CheckSumMismatch,
```
Lipt0nas, line 50:
```suggestion
    ChecksumMismatch,
```

drop-transfer/src/ws/server/v4.rs
```diff
@@ -651,6 +656,20 @@ impl handler::Downloader for Downloader {
...
+            let file = std::fs::File::open(&path.0)?;
+            let csum = file::checksum(&mut io::BufReader::new(file))?;
+            crate::Result::Ok(csum)
+        })?;
+
+        if self.full_csum.get().await == csum {
+            Ok(())
+        } else {
+            Err(crate::Error::CheckSumMismatch)
+        }
```
Lipt0nas, line 671: Mega minor: To fall in-line with the refactor in `ws/server/mod.rs`, but can ignore this one

```suggestion
        if self.full_csum.get().await != csum {
            return Err(crate::Error::CheckSumMismatch)
        }
        
        Ok(())
```

drop-core/src/status.rs
matszczygiel: Applied

drop-transfer/src/ws/server/v4.rs
matszczygiel: Applied

Lipt0nas (approved):

PR #92, opened 2023-06-22
[auto retries] address transfer request mismatch

matszczygiel:


drop-transfer/src/ws/server/mod.rs
```diff
@@ -310,51 +310,58 @@ async fn handle_client(
...
+        };
+
+        let current_info = xfer.storage_info();
+        if let Some(mut existing_info) = existing_info {
+            // Check if the transfer matches
+            anyhow::ensure!(
+                current_info.peer == existing_info.peer,
+                "Transfer {} resume. Peers do not match",
+                xfer.id()
+            );
```
LukasPukenis, line 340: This whole close is very difficult to read and validate that it is correct, could you split it into multiple, smaller parts?

matszczygiel: Yes, let me try

matszczygiel: Refactored a bit. Can you take a look?

drop-transfer/src/ws/server/mod.rs
```diff
@@ -780,3 +712,111 @@ fn move_tmp_to_dst(
...
+
+        state
+            .event_tx
+            .send(Event::RequestReceived(xfer.clone()))
+            .await
+            .expect("Failed to notify receiving peer!");
+    }
+
+    Ok(())
+}
```
LukasPukenis, line 768: It's clearer now to see that when we receive a transfer we match it with what exists in the database, and if it's _exactly_ the same we automatically initiate the resume while omitting `TransferRequested` event.

Two points that stand out:
- we use the unbounded channel, thus if our transfer contains 1000 files for example, we will initiate 1000 parallel downloads with no rate-limiting
- Why do we use a channel to communicate transfer tasks but we also have `transfer_manager`, it looks like two detached communication channels. Maybe it's worth wrapping the channel into TransferManager?

matszczygiel: Yes, we do not have a rate limit, but it's the same for the app `download` calls. Actually, the number of files to restart here is exactly what the app allowed on the previous run.
Regarding the `transfer_manager`, the same chancel is stored there. I've extracted it only to not clutter these methods with manager calls which is additionally mutex protected

LukasPukenis: The app at least controls how many `download()` calls it issues. 

Consider where the app tries one by one to download 100 files and all fail. After restarting all 100 will be in flight.
It would be nice to rate-limit to some sane number, like providing a configuration limit with a default of 16 for example.

The problem with that though is that it might race with other API calls given that we queue.

LukasPukenis (approved):
Rate limit comment still applies however the logic was already present, moving that into a ticket instead.

PR #93, opened 2023-06-23
Add support for IPv6

matszczygiel:


drop-transfer/src/ws/client/mod.rs
```diff
@@ -142,7 +142,11 @@ async fn make_request(
     auth: &auth::Context,
     logger: &slog::Logger,
 ) -> Result<(), tungstenite::Error> {
-    let url = format!("ws://{ip}:{}/drop/{version}", drop_config::PORT);
+    let addr = SocketAddr::new(ip, drop_config::PORT);
+
+    let url = format!("ws://{addr}/drop/{version}",);
+
+    debug!(logger, "Making HTTP request: {url}");
```
LukasPukenis, line 149: nitpick: if drop-transfer is agnostic to IPv4/6 maybe it's worth extracting it into FFI layer and pass it down here

LukasPukenis (approved):
very nice PR

drop-transfer/src/ws/client/mod.rs
matszczygiel: Um, sorry but I don't get it. What do you want to move to ffi layer? the URL construction?

LukasPukenis: Address construction

matszczygiel: So you want to pass the `SocketAddr` from the ffi layer. It actually makes sense. This way we would detach `drop-transfer` crate from the concrete port.
Let me do this one in Monday

matszczygiel: @LukasPukenis As it turns out it's not an easy task, because we construct the address from the `Transfer` object which cannot store port information as it is not stable per peer

PR #94, opened 2023-06-26
Fetch preexisting transfer directories on resume

matszczygiel:
So we have this cache of our own created directories in the `TransferManager` but it's stored in RAM and is wiped out on transfer failures.
This PR implements restoring the cache from DB in case the transfer is resumed. It's an alternative to marker files


LukasPukenis:
It's not particularly clear what this does.. The description is empty and no testcase

matszczygiel:
@LukasPukenis I've updated the description

Lipt0nas (approved):
Looks good, but perhaps we should wait for @LukasPukenis to have a look at it after the description update as well

LukasPukenis (approved):

PR #95, opened 2023-06-27
Do not retry rejected files

matszczygiel:


LukasPukenis (approved):
PR looks like it makes sense.
We must have a test site around this though.

matszczygiel:
Yeah, for sure testing is needed

PR #97, opened 2023-06-27
Fix clippy unsafe dereference warning

matszczygiel:


norddrop/src/ffi/mod.rs
```diff
@@ -77,8 +77,11 @@ pub struct norddrop(Mutex<NordDropFFI>);
 ///   }
 /// ]
 /// ```
+///
+/// # Safety
+/// The pointers provided must be valid
```
Lipt0nas, line 82: Since these will modify the resulting `norddrop.h` file, could you also include the modified bindings as well?

matszczygiel: Sure

matszczygiel: Done

Lipt0nas (approved):

PR #98, opened 2023-06-28
Automated retries for android

matszczygiel:
This PR adds support for content URIs on Unix OSes. The FD is removed from `TransferDescriptor` and is fetched from the app via callback.
The callback itself is set by the C `norddrop_set_fd_resolver_callback` method and I exposed it only in the Java bindings since only Android will rely on this.
In order to persist the transfer containing content URIs, we decided to go for storing generalized URIs in the DB so that
`file:///<path>` becomes a URI of an ordinary file and `content:///<identifier>` is an app-defined identifier to be passed down the FD callback.


drop-storage/src/types.rs
```diff
@@ -240,7 +249,10 @@ pub struct OutgoingPath {
     #[serde(serialize_with = "serialize_datetime")]
     pub created_at: NaiveDateTime,
     pub transfer_id: TransferId,
-    pub base_path: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub base_path: Option<PathBuf>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub content_uri: Option<url::Url>,
```
matszczygiel, line 255: This is a JSON API structure. The question is do we change something here regarding the URIs?
I see that even if we decide to not pass raw URIs here the `base_path` can be changed to `full_path` as it might be more useful for apps

test/drop_test/action.py
```diff
@@ -135,23 +131,18 @@ def __init__(self, peers: typing.List[str], path: str):
         self._fd: typing.Any = None
```
LukasPukenis, line 131: This isn't used anymore

test/drop_test/action.py
```diff
@@ -135,23 +131,18 @@ def __init__(self, peers: typing.List[str], path: str):
...
+            )
             UUIDS.append(xfid)
 
             UUIDS_LOCK.release()
 
     def __str__(self):
-        return (
-            f"MultipleNewTransfersWithSameFD({self._peers}, {self._path}, {self._fd})"
-        )
+        return f"MultipleNewTransfersWithSameFD({self._peers}, {self._path})"
```
LukasPukenis, line 145: Isn't it worth just using `NewTransferWithFD` ? This function does nothing wit the `fd`s anymore

test/drop_test/ffi.py
```diff
@@ -173,6 +173,40 @@ def secret(self):
...
 
 
+class FdResolver:
+    def __init__(self):
+        self._files = []
+        self._cached = {}
+
+    def callback(self, ctx, uri):
+        uri: str = uri.decode("utf-8")
+        path = uri.removeprefix("content://")
```
LukasPukenis, line 183: `content://` or `cached` or `new` are entirely this test suite's things. I suggest simplifying the state machine and just speaking using file paths as those are pretty unique and enough for test cases. This way this callback would be very simple by just opening the path and returning a file descriptor.

drop-transfer/src/service.rs
```diff
@@ -82,6 +86,8 @@ impl Service {
                 config,
                 auth: auth.clone(),
                 storage,
+                #[cfg(unix)]
+                fdresolv,
```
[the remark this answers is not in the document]
LukasPukenis: Also, can we maybe expose this function just for UNIX? We need to make our APIs as hard to misuse as possible, so if Windows wouldn't even have a chance to see it in any way

LukasPukenis (approved):
Overall very good effort. I left some comments about simplification which I would like to see if possible

norddrop/src/device/mod.rs
```diff
@@ -631,3 +660,31 @@ async fn open_database(
...
+
+#[cfg(unix)]
+fn crate_fd_callback(
+    logger: slog::Logger,
+    fd_cb: ffi_types::norddrop_fd_cb,
+) -> Arc<drop_transfer::file::FdResolver> {
+    let fd_cb = std::sync::Mutex::new(fd_cb);
+
+    let func = move |uri: &str| {
+        let cstr_uri = format!("{uri}\0").into_bytes();
```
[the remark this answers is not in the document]
matszczygiel: Done

drop-transfer/src/service.rs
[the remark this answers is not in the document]
matszczygiel: Do you mean moving it to a separate function? It would be more prone to misuse then because the callback can't be dynamically switched on down the row. It must be present from the beginning.
On the ffi layer, it's a different story. There we must somehow not bother apps with this callback except Android so as I see it is beneficial to instruct only the Android team to call this additional function

test/drop_test/action.py
matszczygiel: Let me see. It might be a leftover

matszczygiel: Oh, I see now. So this reuses the FD multiple times and the `NewTransferWithFD` creates a new one each time

test/drop_test/ffi.py
matszczygiel: Some test cases resue the FDs so the caching mechanism must be present

test/drop_test/action.py
matszczygiel: Removed

drop-storage/src/types.rs
matszczygiel: So we decided to not change the JSON API

test/drop_test/ffi.py
LukasPukenis: the caching mechanism is fine but it can be as simple as just a hashmap with path as a key and fd as a value and if it's there - it's cached the way I see it

matszczygiel: Yes, it is there. Some tests are refusing FDs and some do not. So I've devised these `cached/` and `new/` to instruct the callback to do the right thing

test/drop_test/action.py
LukasPukenis: A bool flag indicating to cache the fd and reuse it would be nice improvement imho

test/drop_test/ffi.py
LukasPukenis: Oh I see, if it's a callback then it needs to differentiate between cached and non-cached paths. Makes sense

test/drop_test/action.py
matszczygiel: Done

LukasPukenis (approved):

PR #99, opened 2023-06-28
SQLite library clash

Lipt0nas:
Since we don't need to parse the context at any point during the runtime, I have made this into an argument for `norddrop_start`, this should also be less painful for the app teams to implement

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,8 @@
 ---
 * Add IPv6 support
 * Add file rejections via `norddrop_reject_file()`
+* Do not allow the same moose path to be used for app and norddrop trackers
+* Pass app tracker context to `norddrop_start`
 
 ---
 <br>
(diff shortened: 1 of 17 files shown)


matszczygiel:
@Lipt0nas You meant that the argument is nicer for the apps than the config parameter?

Lipt0nas:
The initial idea was to have a callback, where we could request the config from the apps, instead of providing it at the start, I wasn't considering putting it in the config, as it would make the config a little too noisy and since this should be a temporary solution, having it as an argument means that app integrating will have to pay more mind to it and hopefully also remember to not use the same path for the trackers too :D

matszczygiel:
Yeah, so my opinion is that it would be cleaner to put in the `MooseCofnig` as a required field. In that way, we can error on configs missing it - which should give the apps awareness. At the same time, the ABI would be unchanged.
Moreover, after we eventually solve the problem we can silently remove the config param without apps even noticing and thus requiring action - the extra fields are simply ignored by serde deserializer

Lipt0nas:
Good point, I do like the fact that we will not be touching the ABI

matszczygiel (approved):
Nice :+1:

PR #101, opened 2023-06-29
[Main Branch] Switch to rusqlite

Lipt0nas:
decided to do a clean start, since there were a saddening amount of conflicts and it gave me a chance to split up the changes in cleaner commits


drop-storage/src/lib.rs
```diff
@@ -1,630 +1,532 @@
...
 pub struct Storage {
     _logger: Logger,
-    conn: SqlitePool,
+    pool: Pool<SqliteConnectionManager>,
+}
+
+static MIGRATIONS_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/migrations");
+
+lazy_static! {
+    static ref MIGRATIONS: Migrations<'static> = Migrations::from_directory(&MIGRATIONS_DIR)
```
matszczygiel, line 34: I wouldn't say I like it. Why do we use static here? Can't we create the thing at the call site?

drop-storage/src/lib.rs
```diff
@@ -1,630 +1,532 @@
...
 
-    pub async fn insert_transfer(&self, transfer: &TransferInfo) -> Result<()> {
+    pub fn insert_transfer(&self, transfer: &TransferInfo) -> Result<()> {
         let transfer_type_int = match &transfer.files {
             TransferFiles::Incoming(_) => TransferType::Incoming as u32,
             TransferFiles::Outgoing(_) => TransferType::Outgoing as u32,
         };
 
-        let tid = transfer.id.hyphenated();
+        let tid = transfer.id.hyphenated().to_string();
```
matszczygiel, line 63:
```suggestion
        let tid = transfer.id.to_string();
```

drop-storage/src/lib.rs
```diff
@@ -1,630 +1,532 @@
...
-            ON CONFLICT DO NOTHING
-            "#,
-            tid,
-            path.relative_path,
-            path.file_id,
-            path.size,
-        )
-        .execute(conn)
-        .await?;
+        let tid = transfer_id.hyphenated().to_string();
```
matszczygiel, line 96: Same, the `hypenathed()` thing was strictly for sqlx

drop-transfer/examples/udrop.rs
```diff
@@ -58,7 +58,7 @@ async fn listen(
     let mut storage = drop_transfer::StorageDispatch::new(&storage);
 
     while let Some(ev) = rx.recv().await {
-        if let Err(e) = storage.handle_event(&ev).await {
+        if let Err(e) = storage.handle_event(&ev) {
```
matszczygiel, line 61: TBH we might want to use the `tokio::task::block_in_place()` here

matszczygiel:
In general, it's OK. I simply don't like downgrading the code quality and developer experience but that's what we need to do.
It's sad that basically with the migrations it's a one-way ticket and it would be virtually impossible to go back to sqlx eventually

matszczygiel (dismissed):

drop-transfer/examples/udrop.rs
Lipt0nas: done

drop-storage/src/lib.rs
Lipt0nas: wasn't aware that this was only for sqlx, fixed

drop-storage/src/lib.rs
Lipt0nas: good point, done

drop-storage/src/lib.rs
Lipt0nas: done

matszczygiel (approved):

LukasPukenis:
<3

PR #103, opened 2023-06-30
Automated retries tests

matszczygiel:
I've extracted the `start()` call into the test cases itself. This enables multiple restarts of the peer inside the test cases and thus we can test the retries.
I've also added 4 basic retries tests

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -5304,4 +5304,524 @@ scenarios = [
             ),
         },
     ),
+    Scenario(
+        "scenario29-1",
+        "Send one file to a peer, stop the sender and then start back. Expect automatically restored transfer",
+        {
+            "ren": ActionList(
+                [
+                    action.ConfigureNetwork(),
+                    action.Start("172.20.0.5", dbpath="/tmp/db/29-1-ren.sqlite"),
+                    action.WaitForAnotherPeer(),
+                    action.NewTransfer("172.20.0.15", ["/tmp/testfile-big"]),
+                    action.Wait(
+                        event.Queued(
+                            0,
+                            {
+                                event.File(
+                                    FILES["testfile-big"].id,
+                                    "testfile-big",
+                                    10485760,
+                                ),
+                            },
+                        )
+                    ),
+                    action.Wait(event.Start(0, FILES["testfile-big"].id)),
+                    # wait for the initial progress indicating that we start from the beginning
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id, 0)),
+                    # make sure we have received something, so that we have non-empty tmp file
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id)),
+                    action.Stop(),
+                    action.Wait(event.FinishFailedTransfer(0, Error.CANCELED)),
+                    action.WaitForAnotherPeer(),
+                    # start the sender again
+                    action.Start("172.20.0.5", dbpath="/tmp/db/29-1-ren.sqlite"),
+                    action.Wait(
+                        event.Queued(
+                            0,
+                            {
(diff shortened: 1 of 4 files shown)


test/scenarios.py
```diff
@@ -5198,4 +5304,524 @@
...
+                    action.NoEvent(),
+                    action.Stop(),
+                ]
+            ),
+        },
+    ),
+    Scenario(
+        "scenario29-3",
+        "Send three files to a peer, download one, reject one and do nothing with third one. Then stop the sender and then start back. Expect automatically restored transfer, without the rejected and stopped file",
+        {
```
LukasPukenis, line 5523: That is a great testcase

test/scenarios.py
```diff
@@ -5198,4 +5304,524 @@
...
+                        )
+                    ),
+                    action.Wait(
+                        event.FinishFileRejected(0, FILES["testfile-bulk-01"].id, True)
+                    ),
+                    action.Wait(event.Start(0, FILES["testfile-big"].id)),
+                    # wait for the initial progress indicating that we start from the beginning
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id, 0)),
+                    # make sure we have received something, so that we have non-empty tmp file
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id)),
```
LukasPukenis, line 5566: These three might racy, maybe wrap them inside Action.WaitRacy

LukasPukenis (approved):
Great test cases

test/scenarios.py
matszczygiel: Actually, it's not racy. If the order is different then it's a manifestation of a bug

test/scenarios.py
```diff
@@ -5198,4 +5304,524 @@
...
+                        )
+                    ),
+                    action.Wait(event.Start(0, FILES["testfile-big"].id)),
+                    # wait for the initial progress indicating that we start from the beginning
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id, 0)),
+                    # make sure we have received something, so that we have non-empty tmp file
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id)),
+                    action.Stop(),
+                    action.Wait(event.FinishFailedTransfer(0, Error.CANCELED)),
+                    action.WaitForAnotherPeer(),
```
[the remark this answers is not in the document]
matszczygiel: Yeah, that's a good question. So on `stop()` the `Failed` event is fired but with `Cancelled` status. It might be beneficial to change this

test/scenarios.py
LukasPukenis: Could you elaborate? Isn't that each file is being asynchronously started? Does this mean the transfers are initiated one by one?

test/scenarios.py
```diff
@@ -5198,4 +5304,524 @@
...
+                            0,
+                            {
+                                event.File(
+                                    FILES["testfile-big"].id,
+                                    "testfile-big",
+                                    10485760,
+                                ),
+                            },
+                        )
+                    ),
```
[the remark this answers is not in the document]
matszczygiel: Yes, in both cases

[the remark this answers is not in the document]
LukasPukenis: So it's symmetrical, if that's so then it's OK.

test/scenarios.py
matszczygiel: It is started only if it was in flight during the disconnection. In this test case, we only issue download for one file so the events are in order

LukasPukenis: Oh I think I got confused because it's two progress events for the same file

PR #104, opened 2023-07-03
DDoS protection

matszczygiel:


drop-transfer/src/ws/server/mod.rs
```diff
@@ -128,8 +135,14 @@ pub(crate) fn start(
...
             }))
             .and(
-                warp::filters::addr::remote().then(|peer: Option<SocketAddr>| async move {
-                    peer.expect("Transport should use IP addresses")
+                warp::filters::addr::remote().then(move |peer: Option<SocketAddr>| {
+                    let rl = rate_limiter.clone();
+                    async move {
+                        let peer = peer.expect("Transport should use IP addresses");
+
+                        rl.until_key_ready(&peer.ip()).await;
```
LukasPukenis, line 143: Several idas in one comment:
- This looks pretty clean and neat. I imagined we will just drop the connections, as we do not drop, maybe there's a way to check the `governor` that the slot is not ready and a debug log to it so the delays could be tracked down in the logs?

- On the other hand, what do you think about dropping the retries if they reach the limit? In this leaky-bucket case it will potentially leak if we have some erroneous retry loop

- The default size of 5 is way too small IMHO, it would be at least 10x. My main concern is catching erroneous behaviour of our own code where we would spam the peer with infinite retries or something like that.

@Lipt0nas could also give his input :)

matszczygiel: I initially planned to drop the connection and `governor` actually has this feature but it has a flaw that it won't let in any request past the limit. Meaning that if we allow for example 50 requests per sec and somebody spams with 51 per second, no request is going to be handled whatsoever (I imagined that 50 of then still would be handled). Do you think that's OK?

I agree about the limit

matszczygiel: Ok, so I was wrong. As it turns out it will pass this 50 of 51 requests

matszczygiel: Fixed. Now it will respond with 429 when the limit is reached

LukasPukenis (commented):
Just to be sure: it limits on the receiver/server per transfer defaulted to 50 transfers/sec. After that's exceeded TooManyRequests HTTP code is returned and the transfer failed. Is that correct? :)

Now as it returns an error it's safe to actually test it. I am interested in exercising it in two ways:
- when the limit is reached and further transfers are rejected
- when the limit is reached and after a cooldown transfers are allowed again.

What do you think about those cases, are they achievable?

matszczygiel:
@LukasPukenis let me try

matszczygiel:
@LukasPukenis I managed to devise a test case triggering the DDoS protection. It's in the last commit

test/scenarios.py
```diff
@@ -5947,4 +5947,116 @@
...
+                            "172.20.0.5",
+                            {
+                                event.File(
+                                    FILES["testfile-small"].id,
+                                    "testfile-small",
+                                    1048576,
+                                ),
+                            },
+                        )
+                    ),
```
LukasPukenis, line 6052: I think it might help to use different files in transfers to better distinguish the transfers, however, what's weird is that the receiver is made aware of the transfer via `RequestReceived` but then is not made aware of it's failure so we should either send an additional failure event as well or don't emit the event at all.

I would prefer not emitting any event at all in this case as we can flood the receiver with events as a DDoS

matszczygiel: The receiver does not get any event in case of failure. I think you might have misunderstood this as the UUID slots are local to the runner so Stimpy's slot 2 is the same transfer as Ren's slot 3 in this case

LukasPukenis: I mean receiver receives the `RequestReceived` then it responds with TooManyRequests behind the scenes to the sender but the receiver itself is oblivious that the transfer is failed

matszczygiel: True, that's what happens here. Do you suggest something else? I would argue that the current approach is correct

matszczygiel: Ok so from the beginning. If we are a receiver and a peer makes HTTP request to us and the DDoS protection is triggered we do not receive any event. As a user, we basically don't know about the request at all.
The sender however receives a `TransferFailed` event.

Is this approach somehow wrong in your opinion?

LukasPukenis: That sounds very correct and exactly how I would expect it to be. It looks correct now as it actually initiates with 3 transfers instead of 2(i looked only at events), sorry about the confusion.

LukasPukenis (approved):

PR #106, opened 2023-07-03
Database open error handing

Lipt0nas:
Try to open database in memory if path fails, in addition, send off the errors with moose

### DIFF
--- a/norddrop/src/device/mod.rs
+++ b/norddrop/src/device/mod.rs
@@ -607,7 +607,8 @@ fn open_database(
         }
     }
 
-    // If we can't even open the DB in memory, there is nothing else left to do, throw an error
+    // If we can't even open the DB in memory, there is nothing else left to do,
+    // throw an error
     if dbpath == ":memory:" {
         let error = ffi::types::NORDDROP_RES_DB_ERROR;
         let error_msg = "Failed to open in-memory DB";
@@ -620,8 +621,15 @@ fn open_database(
         );
 
         error!(logger, "{}", error_msg);
-        return Err(error);
+        Err(error)
     } else {
+        moose.developer_exception(
+            -1,
+            ffi::types::NORDDROP_RES_DB_ERROR as i32,
+            "Initial DB open failed, recreating".to_string(),
+            "Failed to open database".to_string(),
+            "Database Error".to_string(),
+        );
         // Still problems? Let's try to delete the file, provided it's not in memory
         warn!(logger, "Removing old DB file");
         if let Err(err) = std::fs::remove_file(dbpath) {
--- a/norddrop/src/device/mod.rs
+++ b/norddrop/src/device/mod.rs
@@ -598,31 +598,44 @@ fn open_database(
     for i in 1..=3 {
         match drop_storage::Storage::new(logger.clone(), dbpath) {
             Ok(storage) => return Ok(storage),
-            Err(err) => warn!(logger, "Failed to open DB: {err}, already tried {i} times",),
+            Err(err) => error!(
+                logger,
+                "Failed to open DB at \"{dbpath}\": {err}, already tried {i} times",
+            ),
         }
     }
 
-    // Still problems? Let's try to delete the file
-    warn!(logger, "Removing old DB file");
-    if let Err(err) = std::fs::remove_file(dbpath) {
-        error!(
-            logger,
-            "Failed to opend DB and failed to remove it's file: {err}"
-        );
+    // If we can't even open the DB in memory, there is nothing else left to do, throw an error
+    if dbpath == ":memory:" {
+        error!(logger, "Failed to open in-memory DB");
         return Err(ffi::types::NORDDROP_RES_DB_ERROR);
     } else {
-        // Inform app that we wiped the old DB file
-        events.dispatch(types::Event::RuntimeError {
-            status: drop_core::Status::DbLost,
-        });
-    };
+        // Still problems? Let's try to delete the file, provided it's not in memory
+        warn!(logger, "Removing old DB file");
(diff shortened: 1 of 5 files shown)


norddrop/src/device/mod.rs
```diff
@@ -593,36 +595,71 @@ fn open_database(
...
     // Try opening the DB 3 times
     for i in 1..=3 {
         match drop_storage::Storage::new(logger.clone(), dbpath) {
             Ok(storage) => return Ok(storage),
-            Err(err) => warn!(logger, "Failed to open DB: {err}, already tried {i} times",),
+            Err(err) => error!(
+                logger,
+                "Failed to open DB at \"{dbpath}\": {err}, already tried {i} times",
+            ),
         }
```
Lipt0nas, line 608: I considered adding a moose call here as well, but i don't think it would give enough value given how much spam it would introduce

LukasPukenis: Not worth logging to moose if we iterate(its probably gonna be the same error multiple times) imho

LukasPukenis: Thinking about it twice - it would be nice to know at least once that this happened as other logs might not be exercised if everything went fine. This way we're left oblivious to the fact that the database failed but it was properly flushed and recreated.

Thinking of it - we can emit a moose event here but remove the retry. It could be like this:
- try to use a database
- if fail -> emit moose event(1), try to recreate it 
- if fail recreating -> emit moose event(2) and use `:memory:`
- if fail using `:memory:` -> emit moose event(3) and fail with an error

Lipt0nas: done

norddrop/src/device/mod.rs
```diff
@@ -593,36 +594,79 @@ fn open_database(
...
-            Err(ffi::types::NORDDROP_RES_DB_ERROR)
+        // Final try after cleaning up old DB file
+        match drop_storage::Storage::new(logger.clone(), dbpath) {
+            Ok(storage) => Ok(storage),
+            Err(err) => {
+                let error = ffi::types::NORDDROP_RES_DB_ERROR;
+                let error_msg = format!("Failed to open DB after cleaning up old file: {err}");
+                moose.developer_exception(
+                    -1,
+                    error as i32,
```
matszczygiel, line 662: Isn't moose using the status codes? I don't know if it even matters but this one is just a cast of the ffi error, not a status

Lipt0nas: This is a little different than the other moose calls that we have, as it's generic for all of the trackers, so we use the argument that seems to fit best

norddrop/src/device/mod.rs
LukasPukenis: @Lipt0nas what do you think about my following(after the initial comment) suggestion?

norddrop/src/device/mod.rs
```diff
@@ -593,36 +594,79 @@ fn open_database(
...
+            "Failed to open database".to_string(),
+            "Database Error".to_string(),
+        );
+        // Still problems? Let's try to delete the file, provided it's not in memory
+        warn!(logger, "Removing old DB file");
+        if let Err(err) = std::fs::remove_file(dbpath) {
+            let error_msg = format!("Failed to open DB and failed to remove it's file: {err}");
+            moose.developer_exception(
+                -1,
+                ffi::types::NORDDROP_RES_DB_ERROR as i32,
```
LukasPukenis, line 639: few very minor nitpicks:
- `-1` could be a constant with a constant why it's negative 1
- `ffi::types` could have a conversion to i32

norddrop/src/device/mod.rs
matszczygiel: Ok, I guess it's not very important after all since we have the description passed into the function

norddrop/src/device/mod.rs
Lipt0nas: I left the retrying part in, but otherwise it's essentially what you described, we try to open, if it doesn't work, we try to use `:memory:` and inform of that fact in the description of the event, and if that fails, we emit and event and error out

matszczygiel (dismissed):

norddrop/src/device/mod.rs
Lipt0nas: I could make a different function that uses the `arbitrary_value` and one that doesn't as well, and it's -1, because it's unused, the description of the function on our end is taken from the tracker itself

For the conversion, I thought it's very redundant to have a `Into<i32>` implementation, as the `ff::types` is `#[repr(C)]` with defined numeric values, including a conversion would actually be copy and pasting identical code and changing `as i32` to `.into()`

norddrop/src/device/mod.rs
LukasPukenis: We still have `for i in 1..=3 {...}` present, which I feel is unecessary

norddrop/src/device/mod.rs
LukasPukenis: Fair enough

LukasPukenis (dismissed):
I think the multiple retry logic could go away, no need to repeat the same thing 3 times but overall it's a nice improvement

LukasPukenis:
An additional idea - a test would be nice which would exercise database failure and :memory: usage. This would probably require storing database in root owned path so the runner could not access it and fail but it might be a bit tricky as it's transparent to us

norddrop/src/device/mod.rs
Lipt0nas: removed it

LukasPukenis (approved):

matszczygiel (approved):

PR #108, opened 2023-07-05
add remove_transfer_file() method

matszczygiel:


norddrop/src/device/mod.rs
```diff
@@ -304,6 +304,31 @@ impl NordDropFFI {
...
+            .blocking_lock()
+            .as_ref()
+            .ok_or(ffi::types::NORDDROP_RES_NOT_STARTED)?
+            .remove_transfer_file(transfer_id, &(file_id.into()));
+
+        match res {
+            Ok(()) => Ok(()),
+            Err(drop_transfer::Error::StorageError) => Err(ffi::types::NORDDROP_RES_DB_ERROR),
+            Err(drop_transfer::Error::InvalidArgument) => Err(ffi::types::NORDDROP_RES_BAD_INPUT),
+            Err(_) => Err(ffi::types::NORDDROP_RES_ERROR),
```
LukasPukenis, line 328: nice marshalling of errors

drop-storage/src/lib.rs
```diff
@@ -842,6 +842,47 @@ impl Storage {
...
+                    AND id IN(SELECT path_id FROM incoming_path_reject_states)
+            "#,
+            )?
+            .execute(params![tid, file_id])?;
+
+        if count > 0 {
+            Ok(Some(()))
+        } else {
+            Ok(None)
+        }
```
LukasPukenis, line 883:
```suggestion
        match count {
          0 => Ok(None)
          1 => Ok(Some(())),
          _ => {
            warn!(self.logger, "Deleted a file from both outgoing and incoming paths");
            Ok(Some(()))
          }
        }
```

drop-storage/src/lib.rs
```diff
@@ -1172,4 +1213,103 @@ mod tests {
...
+
+        let transfers = storage.transfers_since(0).unwrap();
+        assert_eq!(transfers.len(), 1);
+
+        let paths = match &transfers[0].transfer_type {
+            DbTransferType::Outgoing(out) => out,
+            _ => panic!("Unexpected transfer type"),
+        };
+        assert_eq!(paths.len(), 1); // 1 since we removed one of them
+        assert_eq!(paths[0].file_id, "id4");
```
LukasPukenis, line 1264: minor: it's worth adding an `assert_eq!` before removal and after to be sure it was really added and then removed :)

LukasPukenis (commented):
an integration test is missing to make this complete, but the code looks solid

matszczygiel:
@LukasPukenis there is no integration test but the functionality is exercised in the unit tests

drop-storage/src/lib.rs
matszczygiel: Done

drop-storage/src/lib.rs
matszczygiel: Done

LukasPukenis:
Unit tests are fine however integration tests are additional safety as they check end-to-end

matszczygiel:
Is blocked by https://github.com/NordSecurity/libdrop/pull/109

LukasPukenis (commented):
It looks like it's good, few ideas:
- Could we make a symmetric test for the receiver as well?
- Tests could have JSON API comparison on both sides to ensure that it's really a local effect

matszczygiel:
@LukasPukenis yeah, good suggestions :+1:

test/scenarios.py
```diff
@@ -5103,4 +5103,252 @@
             ),
         },
     ),
+    Scenario(
+        "scenario29-1",
+        "Remove file on sending side. Expect file not being present in the JSON output",
```
LukasPukenis, line 5108: ### 
```suggestion
        "Remove file on sending side. Expect file not being present in the sender JSON output",
```

LukasPukenis (approved):

PR #109, opened 2023-07-05
fix rejection states not being present in JSON API

matszczygiel:

PR #110, opened 2023-07-06
Fxi tempfile collisions across different transfers

matszczygiel:
This PR changes the way the names of temp files are created. It uses the Android approach where the transfer UUID is attached to the name.
This basically disables the resumes across different transfers but that's OK, since those are deprecated anyway in favor of auto retries.
I've also had to modify the resume test cases to happen inside a single transfer - which turned out to be buggy (cases 21-1, 21-2, 21-3)

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -273,6 +273,21 @@ class WaitRacy(Action):
         return f"WaitRacy({', '.join(str(e) for e in self._events)})"
 
 
+class DrainEvents(Action):
+    def __init__(self, count: int):
+        self._count = count
+
+    async def run(self, drop: ffi.Drop):
+        for i in range(0, self._count):
+            e = await drop._events.wait_for_any_event(100, ignore_progress=True)
+
+            if e is None:
+                raise Exception(f"Missing event number {i} while draining")
+
+    def __str__(self):
+        return f"DrainEvents({self._count})"
+
+
 class NoEvent(Action):
     def __init__(self, duration: int = 6):
         self._duration = duration
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -9,6 +9,7 @@ import subprocess
 import typing
 import json
 import time
+import glob
 
 from . import event, ffi
 from .logger import logger
@@ -360,11 +361,14 @@ class Stop(Action):
 
 
 class ModifyFile(Action):
-    def __init__(self, file: str):
-        self._file = file
+    def __init__(self, file_glob: str):
+        self._file = file_glob
 
     async def run(self, drop: ffi.Drop):
-        with open(self._file, "a") as f:
+        file_list = glob.glob(self._file)
+        file = file_list[0]
+
+        with open(file, "a") as f:
             f.write("42")
 
     def __str__(self):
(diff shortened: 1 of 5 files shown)


matszczygiel:
Is blocked by https://github.com/NordSecurity/libdrop/pull/109

matszczygiel:
The integration tests do not pass because of the pause/resume race bug

test/scenarios.py
```diff
@@ -4247,57 +4179,48 @@
...
-                                ),
-                                event.File(
-                                    FILES["nested/big/testfile-02"].id,
-                                    "nested/big/testfile-02",
-                                    10485760,
-                                ),
-                            },
-                        )
+                    action.CancelTransferFile(0, FILES["nested/big/testfile-01"].id),
+                    action.CancelTransferFile(0, FILES["nested/big/testfile-02"].id),
```
LukasPukenis, line 4183: This whole testsuite is different mainly because it actually cancels the file instead of the transfer request as it did previously, what is the reason of not leaving the old test and making a new one?

test/drop_test/action.py
```diff
@@ -273,6 +274,21 @@ def __str__(self):
         return f"WaitRacy({', '.join(str(e) for e in self._events)})"
 
 
+class DrainEvents(Action):
+    def __init__(self, count: int):
+        self._count = count
+
+    async def run(self, drop: ffi.Drop):
+        for i in range(0, self._count):
+            e = await drop._events.wait_for_any_event(100, ignore_progress=True)
```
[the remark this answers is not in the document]
matszczygiel: High enough and round. It's hand waving but it's common in the `wait()` actions throughout the test framework

test/scenarios.py
matszczygiel: The old test exercised the resume across the transfer. The new exercises resume within a single transfer.
With per transfer temporaries, by definition, the resume across the transfers would not work

LukasPukenis: It makes sense, canceling a transfer is basically finalizing it and no resume should happen there, yay!

LukasPukenis (commented):
I have some concerns:
- globbing, by using glob patterns instead of exact paths we make it more ambiguous and our actions become not that nice anymore as they are ambiguous, they do multiple actions implicitly instead of verbosely stated.
- DrainEvents is a one-time use function just to drain evens so we would not fail the `NoEvents` later on. We can use `WaitRacy` for all events we expect to happen

Can we improve these somehow?

matszczygiel:
@LukasPukenis let me look into it :+1:

matszczygiel:
@LukasPukenis the DrainEvents() it's actually needed because we cannot predict the events, meaning we wait for two finished events and the final path field is undefined in both of them, so `WaitRacy` also does not work here.
Globing is just a quick way to go and works, it's not perfect but how much time do we want to spend coding the test framework

test/drop_test/action.py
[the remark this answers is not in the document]
LukasPukenis: Re discussion: to make it proper we would need to store file IDs in the table and refer to that. Using globbing is much faster and also only used in one test case so it's pretty safe. Still worth improving in the near future.

LukasPukenis (approved):

PR #114, opened 2023-07-10
Delete temporary files when purging transfers from the database

matszczygiel:


matszczygiel:
Is blocked by https://github.com/NordSecurity/libdrop/pull/110

drop-storage/src/lib.rs
```diff
@@ -897,6 +897,37 @@ impl Storage {
...
+
+        let conn = self.pool.get()?;
+
+        let out = conn
+            .prepare(
+                r#"
+                SELECT DISTINCT path_hash, base_dir
+                FROM incoming_paths ip
+                INNER JOIN incoming_path_started_states ipss ON ip.id = ipss.path_id 
+                WHERE transfer_id = ?1
```
LukasPukenis, line 917: - DISTINCT usage gives me questions. We might have multiple entries thus it will give us one, but in that case I think LIMIT 1 is a tiny bit more understandable. 

- The code fetches temp file locations for transfers that have a started state but it doesn't care if they have any other state, for example, completed transfers should not have a temporary file left over. TLDR for simplicity - why not return without any join?

drop-transfer/src/ws/server/v4.rs
```diff
@@ -539,6 +539,34 @@ impl handler::HandlerLoop for HandlerLoop<'_> {
             .expect("Event channel should always be open");
     }
 
+    async fn finalize_success(self) {
```
LukasPukenis, line 542: Could you correct me if my understanding is right:

The `finalize_failure` is called if the handler received any kind of error and `finalize_success` is called if the handler didn't receive any errors. For example, network failure would issue `finalize_failure` and if the file was properly downloaded or canceled then it would be `finalize_success`.

We're relying on the fact that we only have a temporary file left after canceled or unsuccessful task completion?

matszczygiel: The success is called when the transfer is canceled - meaning peers are not interested in continuing it.
Failure is called in any other cases, mainly per transfer error conditions.

The temps are left in either case, we only delete them when success is called because is case of failure we expect the transfer to be resumed

LukasPukenis: It would be nice to have a central place where we would be explicit about what happens to the files when certain conditions are met.

In this case, we know that the temp file will be deleted in the case of cancel and deleted in a case of success(which will not find the temp file and would fail). I think an improvement could be:
- handler returning some Enum about what to do with the temp file
- handlers themselves deleting the file when they expect it to be done

matszczygiel: The handler would not fail because we catch the error and handle it properly https://github.com/NordSecurity/libdrop/pull/114/files#diff-fb1fb25ba74414819c4a03c6bc3928d7d57a23ba18feb72833d09b8125058e8cR590

matszczygiel: The logic is as follows:
* we fetch all possible temp file locations from the DB
* on transfer cancel we try to delete all potential temp files, if the deletion fails (because the file was successfully transferred, or the user deleted it, or whatever), we log the info and continue

drop-storage/src/lib.rs
LukasPukenis: As per discussion: there can be multiple temp files because download calls might be canceled and another destination used.

drop-transfer/src/ws/server/v4.rs
LukasPukenis: Re discussion: The failsafe is for successful transfers and not for the files which was the confusion point. In case the transfer was successful but the file handlers didn't clean up properly this is going to clean-up.

LukasPukenis (approved):

LukasPukenis (dismissed):

PR #120, opened 2023-07-11
tests: add db file without permissions case

matszczygiel:


test/scenarios.py
```diff
@@ -4693,6 +4693,82 @@
...
+    Scenario(
+        "scenario26-2",
+        "Provide database with insufficient permissions, expect file share to work",
+        {
+            "ren": ActionList(
+                [
+                    action.DropPrivileges(),
+                    action.Start(
+                        "172.20.0.5",
+                        dbpath="/tmp/no-permissions/no-access-db.sqlite",
```
LukasPukenis, line 4705: Maybe `/root/no-access-db.sqlite`?

LukasPukenis (dismissed):

test/scenarios.py
matszczygiel: Let me explore this tomorrow :+1:

matszczygiel: Done, Also changed one more test case where the `/tmp/no-permissions` was used

LukasPukenis (approved):

PR #121, opened 2023-07-12
Moose file layer

Lipt0nas:


drop-analytics/src/file_impl.rs
```diff
@@ -0,0 +1,199 @@
...
+            app_version,
+            prod,
+        }
+    }
+
+    fn write_event(&self, event: MooseEventType) -> Result<(), std::io::Error> {
+        let data = if !Path::new(&self.event_path).exists() {
+            "[]".to_string()
+        } else {
+            std::fs::read_to_string(&self.event_path)?
```
matszczygiel, line 75: I would use the file reader directly here
```rust
let file = File::open(path)?;
let file = BufReader::new(file):
let events: Vec<MooseEventType> = serde_json::from_reader(file)?;
```

matszczygiel:
Reading the whole file and writing on each event is way too overkill IMHO. The better approach would be to just append the new JSON objects to the existing file, even without wrapping them in JSON array and reading them in a loop for inspection until the EOF.

The rest seems OK

Lipt0nas:
Initially I wanted to not to re-parse the whole file every single event, but since this file layer is only meant to be used in tests, that have small contained scenarios, and since we don't emit a lot of events to begin with, I deemed this to be a fair trade of it being inefficient, but working out of the box for verification in Python, using the same method as we do for storage, if you think that this is not a good solution, then I'll get rid of the re-parsing

matszczygiel:
That's a fair point that it's going to be used in the tests only.
You should check if it compiles under the `default` and `moose` features, it seems to me that you might be missing some `#[cfg()]` gates

matszczygiel (dismissed):

drop-analytics/src/lib.rs
```diff
@@ -1,20 +1,28 @@
 #[cfg(feature = "moose")]
 mod moose_impl;
 
+mod file_impl;
```
matszczygiel, line 4: I mean here `#[cfg(feature = "moose_file")]`

Lipt0nas: the file implementation, just like the mock one, doesn't use actual moose, it's a self-contained implementation, don't see why it shouldn't compile here

matszczygiel: Oh yeah, you're right. It should compile.
Nonetheless, I still think it's better to gate the implementation under the feature flag so that the code is not even present in the production build

drop-analytics/src/file_impl.rs
```diff
@@ -0,0 +1,199 @@
...
+        };
+
+        let mut events: Vec<MooseEventType> = serde_json::from_str(&data)?;
+        events.push(event);
+
+        File::create(&self.event_path)?.write_all(serde_json::to_string(&events)?.as_bytes())?;
+
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
Lipt0nas: Good point, will do

drop-analytics/src/lib.rs
```diff
@@ -81,8 +89,12 @@ fn create(
         let moose = moose_impl::MooseImpl::new(logger, event_path, app_version, prod)?;
         Ok(Arc::new(moose))
     }
+    #[cfg(feature = "moose_file")]
```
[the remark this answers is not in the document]
Lipt0nas: I would argue against this, in it's current state, enabling events would not be optimal at all, as the file is constantly rewritten and reparsed, which would definitely require for this to be done differently, in other scenarios, one should just use the actual moose implementation, if they want to inspect the events

test/scenarios.py
```diff
@@ -81,6 +81,58 @@
...
+                            "app_version": "*",
+                            "prod": false
+                        }""",
+                        """{
+                            "type": "batch",
+                            "phase": "start",
+                            "transfer_id": "*",
+                            "info": {
+                                "mime_type_list": "unknown",
+                                "extension_list": "unknown",
```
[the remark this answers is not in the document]
Lipt0nas: Was planning to add more tests with a different PR, that tackles this alongside the database related events, this is only a starter :)

drop-analytics/src/lib.rs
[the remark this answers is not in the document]
matszczygiel: I agree with @Lipt0nas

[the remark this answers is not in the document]
LukasPukenis: lets quickly discuss this during tomorrows sync

test/scenarios.py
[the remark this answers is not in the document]
LukasPukenis: Different PR is perfectly fine but I think it can be easily forgotten :)

drop-analytics/src/lib.rs
```diff
@@ -110,3 +123,30 @@ pub fn init_moose(
...
 }
+
+#[allow(unused_variables)]
+pub fn moose_file(
+    logger: Logger,
+    event_path: String,
+    app_version: String,
+    prod: bool,
+) -> Arc<dyn Moose> {
+    #[cfg(feature = "moose_file")]
```
matszczygiel, line 134:
```suggestion
#[cfg(feature = "moose_file")]
pub fn moose_file(
    logger: Logger,
    event_path: String,
    app_version: String,
    prod: bool,
) -> Arc<dyn Moose> {
```

matszczygiel: This function is probably not even necessary, you can inline it since it consists of a single statement

Lipt0nas: makes sense

matszczygiel (approved):

PR #122, opened 2023-07-12
Lukasp/test download after canceled transfer

LukasPukenis:


test/scenarios.py
```diff
@@ -1231,9 +1231,68 @@
...
         },
     ),
+    Scenario(
+        "scenario4-11",
+        "Send a request with one file, cancel the request from the sender immediately, expect the other side not to be able to issue download calls",
+        {
+            "ren": ActionList(
+                [
+                    action.Start("172.20.0.5"),
+                    action.ConfigureNetwork(),
```
matszczygiel, line 1241: The network configuration is not needed and will only slow things down

test/scenarios.py
```diff
@@ -1231,9 +1231,68 @@
...
+                    action.CancelTransferRequest(0),
+                    action.Wait(event.FinishTransferCanceled(0, False)),
+                    action.NoEvent(),
+                    action.Stop(),
+                ]
+            ),
+            "stimpy": ActionList(
+                [
+                    action.Start("172.20.0.15"),
+                    action.ConfigureNetwork(),
```
matszczygiel, line 1264: Same here

test/scenarios.py
```diff
@@ -1231,9 +1231,124 @@
...
+            "stimpy": ActionList(
+                [
+                    action.Sleep(8),
+                    action.Start("172.20.0.15"),
+                    action.NoEvent(),
+                    action.Stop(),
+                ]
+            ),
+        },
+    ),
```
LukasPukenis, line 1324: @matszczygiel This test fails even though I would expect it to pass, the Queued event is not sent while there's no connection to the peer.

Queued should mean that the file path was ok, files were traversed and the transfer is OK, in this case, it's never emitted

test/scenarios.py
```diff
@@ -1231,9 +1231,124 @@
...
```
LukasPukenis, line 1348: This test fails on the sender side with a `BadTransfer` :
```
test-ren-1     | 2023-07-12 15:23:23,786:global:DEBUG:Action threw an exception: Unexpected event:
test-ren-1     | FinishFailedTransfer(transfer=614fc4f9-9482-47d3-a2c8-4c2f22e8975a (slot: 0), status=7, os_err=None)
test-ren-1     | while looking for:
test-ren-1     | FinishTransferCanceled(transfer=614fc4f9-9482-47d3-a2c8-4c2f22e8975a (slot: 0), by_peer=False)
```

Which is not what happened.

@matszczygiel it looks like the logic depends on the connection to the peer. Could you take a look into these testcases?

matszczygiel: We definitely have a bug here, I discovered it recently, which prevents from canceling when the peer is offline

test/scenarios.py
matszczygiel: The basic check is done in the `new_transfer()` call. `Queued` means that we have the connection working

LukasPukenis: I like thinking about Queued so it would mean "the paths and information you provided are OK, your provided IP is also in the right format, here are the files I aggregated". Then use the Started event to indicate the action going on. Combined with FILE-393 the API user would always know which transfer is going to be retried. What is your take on this?

matszczygiel: The IP, files, and paths are checked in the `new_transfer()` call so if there is a problem the user will get the error from the function call. So there is really no way to get success from `new_transfer()` and not receive `Queued` at the same time, so emitting it would be redundant. However, I do agree that the app might want to know the details about the transfer immediately since the `new_transfer()` does not report this kind of information.
The second question is: do we want to inform the app that we have a connection active or not?

In general, I think you might be right about emitting `Queued` always, but I'm not sure if leaving apps blind about the connection state is a good idea - maybe we can introduce another event?

LukasPukenis: If the user adds a directory and sends to an offline peer he will be oblivious to the files aggregated until the user appears online. He could query that data from the storage though so it's not that big of a deal.

However as we have `Started` events, should we even introduce the third event in between? 
Queued - all good, here what's going to be done
Started - here's what's happening

I think it even works out nicely in automated retries - we would emit Queued events for all the transfers we would retry

PR #123, opened 2023-07-13
split Transfer type into IncomingTransfer and OutgoingTransfer types

matszczygiel:
The title says it all. Enjoy :)

### DIFF
--- a/drop-transfer/src/file/mod.rs
+++ b/drop-transfer/src/file/mod.rs
@@ -24,18 +24,41 @@ use crate::{utils::Hidden, Error};
 pub type FdResolver = dyn Fn(&str) -> Option<RawFd> + Send + Sync;
 
 const HEADER_SIZE: usize = 1024;
+const UNKNOWN_STR: &str = "unknown";
 
-#[allow(clippy::large_enum_variant)]
+pub trait File {
+    fn id(&self) -> &FileId;
+    fn subpath(&self) -> &FileSubPath;
+    fn size(&self) -> u64;
+    fn mime_type(&self) -> &str;
+
+    fn info(&self) -> FileInfo {
+        FileInfo {
+            mime_type: self.mime_type().to_string(),
+            extension: self
+                .subpath()
+                .extension()
+                .unwrap_or(UNKNOWN_STR)
+                .to_string(),
+            size_kb: (self.size() as f64 / 1024.0).ceil() as i32,
+        }
+    }
+}
+
 #[derive(Clone, Debug)]
-pub enum FileKind {
-    FileToSend {
-        meta: Hidden<fs::Metadata>,
-        source: FileSource,
-        mime_type: Option<Hidden<String>>,
-    },
-    FileToRecv {
-        size: u64,
-    },
+pub struct FileToSend {
+    file_id: FileId,
+    subpath: FileSubPath,
+    meta: Hidden<fs::Metadata>,
+    pub(crate) source: FileSource,
+    mime_type: Option<Hidden<String>>,
+}
+
+#[derive(Clone, Debug)]
+pub struct FileToRecv {
+    file_id: FileId,
+    subpath: FileSubPath,
+    size: u64,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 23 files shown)


drop-transfer/src/file/mod.rs
```diff
@@ -202,58 +221,59 @@ impl File {
...
     /// Calculate sha2 of a file. This is a blocking operation
     pub(crate) fn checksum(&self, limit: u64) -> crate::Result<[u8; 32]> {
-        let reader = match &self.kind {
-            FileKind::FileToSend { source, .. } => reader::open(source)?,
-            _ => return Err(Error::BadFile),
-        };
-
+        let reader = reader::open(&self.source)?;
         let mut reader = io::BufReader::new(reader).take(limit);
         let csum = checksum(&mut reader)?;
```
LukasPukenis, line 278: Can we be sure BadFile is still being returned in such a case?

drop-transfer/src/file/mod.rs
```diff
@@ -202,58 +221,59 @@ impl File {
...
-            FileKind::FileToSend { meta, source, .. } => {
-                let mut reader = reader::open(source)?;
-                reader.seek(io::SeekFrom::Start(offset))?;
-                FileReader::new(reader, meta.0.clone())
-            }
-            _ => Err(Error::BadFile),
-        }
+        let mut reader = reader::open(&self.source)?;
+        reader.seek(io::SeekFrom::Start(offset))?;
+        FileReader::new(reader, self.meta.0.clone())
```
LukasPukenis, line 271: BadFile is not gone in this case, what is the final error happening in case of open failure?

LukasPukenis (commented):
A worthy improvement in my eyes, if CI is green I am happy :)

drop-transfer/src/file/mod.rs
```diff
@@ -48,65 +71,65 @@ pub enum FileSource {
...
-                .path()
-                .strip_prefix(parent)
-                .map_err(|err| crate::Error::BadPath(err.to_string()))?;
+    fn size(&self) -> u64 {
+        self.size
+    }
 
-            let subpath = FileSubPath::from_path(subpath)?;
+    fn mime_type(&self) -> &str {
+        ""
```
Lipt0nas, line 109: we should also return unknown here

drop-transfer/src/file/mod.rs
matszczygiel: `BadFile` was returned when trying to open a file to be received. Now with strong typing, there is no way one could open a wrong file, so the error is gone

drop-transfer/src/file/mod.rs
matszczygiel: TBH it is the same as before the refactor. A file to be received was reporting an empty string and a file to be sent reported mime or "unknown".
Here it's the same. Should I change this one?

drop-transfer/src/file/mod.rs
matszczygiel: disk IO errors, just like before when we passed the `BadFile` check

drop-transfer/src/file/mod.rs
Lipt0nas: since we discussed constructing the file info for received files, I think it makes sense to use the same logic for both sides

matszczygiel: In general yes, but with the mime itself, it's a bit different I guess. The thing is you always report "unknown" from receiving side so at least 50% of reported mimes would be "unknown" and the downside is that you can't really tell then in which % of cases the mime is truly unknown (on the sender side).

drop-transfer/src/file/mod.rs
LukasPukenis: Thanks

drop-transfer/src/file/mod.rs
LukasPukenis: Nice!

drop-transfer/src/file/mod.rs
Lipt0nas: It's a fair point, let's then leave it as it is right and decide if we want to change it when the tracker gets updated

LukasPukenis (approved):

PR #124, opened 2023-07-17
restore optional transfer descriptor FD field

matszczygiel:


Lipt0nas (approved):
Nice fix on the tests as well :+1:

PR #125, opened 2023-07-18
Update tests to check for database creation recovery

Lipt0nas:


test/scenarios.py
```diff
@@ -4728,11 +4728,84 @@
...
+                            "result": 0,
+                            "app_version": "*",
+                            "prod": false
+                        }""",
+                        ]
+                    ),
+                ]
+            ),
+        },
+        dbpath="/tmp/no-permissions",
```
matszczygiel, line 4807: This one is deleted on `dev` I guess. Use `/root`

test/scenarios.py
```diff
@@ -4728,11 +4728,84 @@
...
         },
         dbpath="/tmp/db/26-corrupted.sqlite",
     ),
+    Scenario(
+        "scenario26-1",
+        "Test if the instance can recover when file is innacessible",
+        {
+            "ren": ActionList(
+                [
+                    action.NoEvent(),
```
matszczygiel, line 4768: You're missing the `DropPrivileges` actions. The runners by default run as a root

Lipt0nas:
I see that the tests are quite different here than what is currently in `dev`, maybe it would be best to merge this when `main` is merged into `dev`?

matszczygiel:
You can wait, but I don't see a point. Let's use `/root/db.sqlite` as in dev and we will free ourselves from the hassle to remember to change or merge later

PR #130, opened 2023-07-25
fix relative path not being inserted into DB

matszczygiel:
Tested with udrop


Lipt0nas (approved):
And a test as well, very nice

PR #132, opened 2023-07-26
Update tests after automated retries

Lipt0nas:
Notable changes:
scenario4-7,9,13 Deleted (sender side rejections)
scenario10-1,2 Disabled for now, since they should be relevant again after implementing events for transfer resumes/stops
scenario10-3 Deleted as it's no longer relevant

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -249,18 +249,6 @@ scenarios = [
                             }
                         }""",
                             """{
-                            "type": "file",
-                            "phase": "end",
-                            "result": 1,
-                            "transfer_id": "*",
-                            "transfer_time": "*",
-                            "info": {
-                                "mime_type": "unknown",
-                                "extension": "none",
-                                "size_kb": 10240
-                            }
-                        }""",
-                            """{
                             "type": "init",
                             "phase": "end",
                             "result": 0,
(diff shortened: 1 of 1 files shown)


test/scenarios.py
```diff
@@ -247,18 +247,6 @@
                                 "extension": "none",
                                 "size_kb": 10240
                             }
-                        }""",
```
matszczygiel, line 250: Nice :+1:

test/scenarios.py
```diff
@@ -1430,30 +1275,6 @@
             ),
         },
     ),
-    Scenario(
```
matszczygiel, line 1433: It's not entirely irrelevant. We might actually want to test this behavior

test/scenarios.py
```diff
@@ -2169,152 +2002,132 @@
...
-                            0,
-                            Error.CANCELED,
-                        )
-                    ),
-                    action.NoEvent(),
-                ]
-            ),
-        },
-    ),
-    Scenario(
```
matszczygiel, line 2297: In this one we just expect `Queued` and no further events

test/scenarios.py
```diff
@@ -5360,10 +5163,10 @@
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
                     action.RejectTransferFile(0, FILES["testfile-big"].id),
                     action.Wait(
-                        event.FinishFileCanceled(0, FILES["testfile-big"].id, False)
+                        event.FinishFileRejected(0, FILES["testfile-big"].id, False)
                     ),
                     action.Wait(
-                        event.FinishFileRejected(0, FILES["testfile-big"].id, False)
+                        event.FinishFileCanceled(0, FILES["testfile-big"].id, False)
                     ),
```
matszczygiel, line 5170: Let's use wait racy, because the order of this events is more or less undefined

test/scenarios.py
```diff
@@ -5452,10 +5255,10 @@
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
                     action.RejectTransferFile(0, FILES["testfile-big"].id),
                     action.Wait(
-                        event.FinishFileCanceled(0, FILES["testfile-big"].id, False)
+                        event.FinishFileRejected(0, FILES["testfile-big"].id, False)
                     ),
                     action.Wait(
-                        event.FinishFileRejected(0, FILES["testfile-big"].id, False)
+                        event.FinishFileCanceled(0, FILES["testfile-big"].id, False)
```
matszczygiel, line 5261: Same with wait racy

test/scenarios.py
```diff
@@ -5782,22 +5585,9 @@
...
-                            {
-                                event.File(
-                                    FILES["testfile-big"].id,
-                                    "testfile-big",
-                                    10485760,
-                                ),
-                            },
-                        )
-                    ),
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
```
matszczygiel, line 5591: I'm not sure if this is correct. We expect `Start` to be emitted only once I guess.
We don't have `Paused` event to matchup with `Start`.

test/scenarios.py
```diff
@@ -5834,13 +5624,6 @@
...
                     ),
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
-                    action.Wait(
-                        event.FinishFailedTransfer(
-                            0,
-                            Error.WS_SERVER,
-                            ignore_os=True,
-                        )
-                    ),
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
```
matszczygiel, line 5627: Same here I would say that the second `Start` is not desired

test/scenarios.py
```diff
@@ -5885,21 +5668,6 @@
...
-                            {
-                                event.File(
-                                    FILES["testfile-big"].id,
-                                    "testfile-big",
-                                    10485760,
-                                ),
-                            },
-                        )
-                    ),
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
```
matszczygiel, line 5671: Same with second `Start`

matszczygiel:
Great job. Am I right that the commented-out tests will be handled in the second run?

test/scenarios.py
Lipt0nas: Then i guess this is a bug in the implementation if we're shooting out start events every time

matszczygiel: Yeah, it might be

Lipt0nas:
The two disabled tests expect to get an error once the other sides goes dark, the idea was to repurpose them when we implement the resume/pause events, right now i don't think they would serve a purpose

matszczygiel (approved):
Regardless of Started events, I think it can be safely merged

PR #134, opened 2023-07-26
clear in_flight file state upon download finish

matszczygiel:


Lipt0nas (approved):
Niiiice, two more tests are green once again :partying_face:

PR #135, opened 2023-07-26
Additional retry tests

Lipt0nas:
Tests added so far:
- Start a transfer to offline peer, restart both, expect transfer to happen
- Start a transfer to offline peer, reject transfer on sender side, restart both, expect events to be synced and the transfer to not happen
- Start a transfer to offline peer, cancel transfer on sender side, restart both, expect events to be synced and the transfer not to happen
- Complete a normal transfer, restart both sides, check that nothing gets resumed
- Send a transfer to a peer, sender goes offline, receiver accepts while sender is offline, sender goes online, expect transfer to happen properly
- Start a transfer to offline peer, reject transfer on receiver side, restart both, expect events to be synced and the transfer to not happen
- Start a transfer to offline peer, cancel transfer on receiver side, restart both, expect events to be synced and the transfer not to happen

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -6289,6 +6289,114 @@ scenarios = [
         },
     ),
     Scenario(
+        "scenario29-12",
+        "Send a transfer request, wait for it to arrive, stop the sender, reject on the receiver, start the sender, expect the transfer to not be resumed",
+        {
+            "ren": ActionList(
+                [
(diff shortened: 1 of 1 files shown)


matszczygiel (approved):
Nice :+1: I assume 10-3 was just redundant and worth removing

Lipt0nas:
yeah, 10-3 was a carbon copy of 5

PR #136, opened 2023-07-27
fix: sometimes the progress event is reported after cancelation

matszczygiel:


Lipt0nas (approved):
Very nice!

Lipt0nas:
Oh, didn't see this was still in draft :sweat_smile:

matszczygiel:
Yeah :) some tests are failing

Lipt0nas:
I tried the tests locally, only the 24-4 was failing, which was because we were checking for specific order, fixed it in a PR

matszczygiel:
The 29-4 is failing but it's not reproducible on a local machine. I will make another PR with a fix. Merging

PR #138, opened 2023-07-28
Temporarily disable checksum tests, fix 27-4 racy events

Lipt0nas:


test/scenarios.py
```diff
@@ -3943,348 +3943,348 @@
...
-                    ),
-                    action.CancelTransferRequest(1),
-                    action.ExpectCancel([1], False),
-                    action.NoEvent(),
-                    action.Stop(),
-                ]
-            ),
-        },
-    ),
+    # Scenario(
```
matszczygiel, line 3946: Let's leave a TODO comment here on why we disable it and when it should be enabled

test/scenarios.py
```diff
@@ -5149,11 +5149,11 @@
...
                     ),
                     action.Wait(event.Start(0, FILES["testfile-big"].id)),
-                    action.Wait(
-                        event.FinishFileCanceled(0, FILES["testfile-big"].id, True)
-                    ),
-                    action.Wait(
-                        event.FinishFileRejected(0, FILES["testfile-big"].id, True)
+                    action.WaitRacy(
+                        [
+                            event.FinishFileRejected(0, FILES["testfile-big"].id, True),
```
matszczygiel, line 5154: The thing is that they probably should be emitted in order. As I think of it more, we might skip the `Canceled` event here like not it tests but in code. It's completely redundant for the API user.
Let's merge this one and think of it later, possibly creating a ticket

matszczygiel (dismissed):

matszczygiel (approved):

PR #142, opened 2023-07-31
Stabilize scenario30

Lipt0nas:
Added a sleep to the receiving side, so that the sending side can fist queue up all of the events and as we can no longer guarantee which of the requests get rejected, we wait for one and only one out of the possible three, the only problem that remains is since we don't know which transfer will get rejected, we don't know which to cancel, so one idea would be to not expect the cancels anymore, but not sure if this is the best solution

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -264,7 +264,6 @@ class Wait(Action):
 class WaitForOneOf(Action):
     def __init__(self, events: typing.List[Event]):
         self._events: typing.List[Event] = events
-        self._found_event = False
 
     async def run(self, drop: ffi.Drop):
         e = await drop._events.wait_for_any_event(100, ignore_progress=True)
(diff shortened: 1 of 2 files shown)


matszczygiel:
Can't we skip canceling the transfers? Just call stop() for both at the end?

test/drop_test/action.py
```diff
@@ -261,6 +261,30 @@ def __str__(self):
...
+    async def run(self, drop: ffi.Drop):
+        e = await drop._events.wait_for_any_event(100, ignore_progress=True)
+
+        if e is None:
+            raise Exception(f"Expected one of {self._events} but got nothing")
+
+        if e not in self._events:
+            raise Exception(f"Expected one of {self._events} but got {e}")
+        else:
+            if self._found_event:
```
matszczygiel, line 278: Is this flag needed?

matszczygiel (approved):

test/drop_test/action.py
Lipt0nas: ah you're right, this only runs once :facepalm:

matszczygiel (approved):

PR #144, opened 2023-07-31
Introduce paused event

matszczygiel:


drop-transfer/src/ws/client/v2.rs
```diff
@@ -265,6 +265,20 @@ impl<const PING: bool> handler::HandlerLoop for HandlerLoop<'_, PING> {
...
+            task.job.abort();
+
+            async move {
+                task.events.pause().await;
+            }
+        });
+
+        futures::future::join_all(tasks).await;
+    }
+
```
Lipt0nas, line 281: Would it make sense to move these identical instances to a common place?

Lipt0nas (approved):
Also a nice catch on the `Panic` event removal

drop-transfer/src/ws/client/v2.rs
matszczygiel: Yes, I would be delighted to do it. But that means a refactoring since the task struct is not unified yet

PR #145, opened 2023-07-31
remove sqlite connection pool

matszczygiel:


drop-storage/src/lib.rs
```diff
@@ -180,7 +180,7 @@ impl Storage {
         remote: Option<sync::FileState>,
         local: Option<sync::FileState>,
     ) -> crate::Result<()> {
-        let mut conn = self.pool.get()?;
+        let mut conn = self.conn.lock().expect("Failed to acquire a lock");
```
LukasPukenis, line 183: We could make our own implementation for `self.pool.get()` to return the connection directly instead from the pool thus reducing the PR, this way there would be no need to copy-paste the error. Not a major problem though.

drop-storage/src/lib.rs
```diff
@@ -976,16 +973,9 @@ impl Storage {
...
+    fn get_outgoing_paths(conn: &Connection, transfer_id: Uuid) -> Result<Vec<OutgoingPath>> {
         let tid = transfer_id.to_string();
 
-        trace!(
-            self.logger,
-            "Fetching outgoing paths for transfer";
-            "transfer_id" => &tid
-        );
-
-        let conn = self.pool.get()?;
```
LukasPukenis, line 988: Why are trace! calls gone?

drop-storage/src/lib.rs
matszczygiel: I'd say that it would be even better to use `tokio::sync::Mutex` (which does not return `Result`) but decided not to for now since it would break the call sites
@Lipt0nas maybe you can do it in your PR? Since you are already breaking the call sites :P

drop-storage/src/lib.rs
matszczygiel: It's a free function now, does not have access to `logger` anymore

matszczygiel: I can change it to take a logger as an argument. What do you think?

LukasPukenis: I think that more logging is helpful when debugging things but you decide if it's worth it to have it here :)

drop-storage/src/lib.rs
Lipt0nas: Yeah, ill change it to tokio's mutex in my PR :+1:

LukasPukenis (approved):

PR #147, opened 2023-07-31
tests: protect event queue with a lock

matszczygiel:


matszczygiel:
Yes. It seems so. The segfault is still there but events are not skipped anymore

Lipt0nas (approved):

LukasPukenis (approved):

PR #149, opened 2023-07-31
[draft] Handle DB errors in drop-transfer crate

Lipt0nas:


drop-storage/Cargo.toml
```diff
@@ -14,12 +14,10 @@ serde = { workspace = true }
 uuid = { workspace = true }
 chrono = { version = "0.4.26", default-features = false, features = ["std"] }
 url = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread"] }
```
matszczygiel, line 17:
```suggestion
tokio = { workspace = true }
```
The `rt-multi-thread` features is already enable in workspace

drop-storage/src/lib.rs
```diff
@@ -62,194 +63,290 @@ impl Storage {
...
-            TransferFiles::Incoming(files) => {
-                trace!(
-                    self.logger,
-                    "Inserting transfer::Incoming files len {}",
-                    files.len()
-                );
-
-                for file in files {
-                    Self::insert_incoming_path(&conn, transfer.id, file)?;
+        let task = || async {
```
matszczygiel, line 66:
```suggestion
        let task = async {
```

drop-storage/src/lib.rs
```diff
@@ -62,194 +63,290 @@ impl Storage {
...
+    pub async fn update_transfer_sync_states(
         &self,
         transfer_id: Uuid,
         remote: Option<sync::TransferState>,
         local: Option<sync::TransferState>,
-    ) -> Result<()> {
-        let mut conn = self.conn.lock().expect("Failed to acquire a lock");
-        let conn = conn.transaction()?;
+    ) {
+        let task = || async {
```
matszczygiel, line 120:
```suggestion
        let task = async {
```

drop-storage/src/lib.rs
```diff
@@ -62,194 +63,290 @@ impl Storage {
...
+        if let Err(e) = task().await {
+            error!(self.logger, "Failed to update transfer sync states"; "error" => %e);
+        }
     }
 
-    pub fn transfer_sync_state(&self, transfer_id: Uuid) -> crate::Result<Option<sync::Transfer>> {
-        let conn = self.conn.lock().expect("Failed to acquire a lock");
-        sync::transfer_state(&conn, transfer_id)
+    pub async fn transfer_sync_state(&self, transfer_id: Uuid) -> Option<sync::Transfer> {
+        let task = || async {
```
matszczygiel, line 142: Same as above. The closure is not needed since the async blocks already encapsulates the task

drop-transfer/src/service.rs
```diff
@@ -129,55 +129,8 @@ impl Service {
...
-            Ok(None) => {
-                warn!(self.logger, "File {file_id} not removed from {transfer_id}");
-                Err(Error::InvalidArgument)
-            }
-            Err(err) => {
-                error!(self.logger, "Failed to remove transfer file: {err}");
-                Err(Error::StorageError)
-            }
-        }
+    pub fn storage(&self) -> &Storage {
```
matszczygiel, line 132: Nice :)

drop-transfer/src/service.rs
```diff
@@ -249,13 +202,14 @@ impl Service {
         if started {
             let file_info = state.xfer.files()[file_id].info();
 
-            let mut task = || {
+            let storage = self.state.storage.clone();
+            let task = || async move {
```
matszczygiel, line 206:
```suggestion
            let task = async {
```

drop-transfer/src/storage_dispatch.rs
```diff
@@ -22,7 +22,7 @@ impl<'a> StorageDispatch<'a> {
         }
     }
 
-    pub fn handle_event(&mut self, event: &crate::Event) -> Result<(), Error> {
+    pub async fn handle_event(&mut self, event: &crate::Event) -> Result<(), Error> {
```
matszczygiel, line 25: Does this function fail anymore?

norddrop/src/device/mod.rs
```diff
@@ -289,19 +283,17 @@ impl NordDropFFI {
...
-                .map_err(|err| {
-                    error!(self.logger, "Failed to get transfers: {:?}", err);
-                    match err {
-                        drop_transfer::Error::StorageError => ffi::types::NORDDROP_RES_DB_ERROR,
-                        _ => ffi::types::NORDDROP_RES_DB_ERROR,
-                    }
-                })
+                .await;
+
+            Ok::<Vec<drop_storage::types::Transfer>, ffi::types::norddrop_result>(transfers)
```
matszczygiel, line 296: Doesn't simple `Ok(transfers)` work?

matszczygiel (dismissed):

drop-transfer/src/storage_dispatch.rs
Lipt0nas: good point, it doesn't anymore :+1:

norddrop/src/device/mod.rs
matszczygiel: Does `Result::Ok(transfers)` work? The `Result` is declared in this file for exactly the `norddrop_result` error type so it should work

LukasPukenis:
It would be great to have more explanatory commit messages in history, instead of:

- `Use Tokio mutex instead of std for storage calls` we could explain what's the problem with regular mutexes(those block entire thread with multiple tokio tasks on them).

- `Handle DB errors in drop-transfer crate` could be expanded with the reasoning that we don't want any failures of the database to interfere with the functionality of the code.

matszczygiel (approved):

PR #150, opened 2023-08-01
tests: fix segfault

matszczygiel:
Looking at the stack traces of core dumps it seems that the crashes are happening because some tokio tasks are still running at the ferry end of test execution and they try to call into Python code which is probably non-existent anymore.
The way to ensure the tasks are closed is to drop the tokio runtime but it only happens in the `norddrop_destory()` call, which btw should be called by the integration at the end of the operation.
So I just run the `norddrop_destory()` at the end of the test execution and it seems to work.

A similar error was happening before we introduced the `Stop()` action in the tests. Then it was the server task crashing the runner which is stopped in the `norddrop_stop()` call. Now it's the client tasks, which are signaled to finish in the `norddrop_stop()` call but are not awaited for.

### DIFF
--- a/test/drop_test/ffi.py
+++ b/test/drop_test/ffi.py
@@ -496,6 +496,12 @@ class Drop:
 
         return ctypes.string_at(version).decode("utf-8")
 
+    def __del__(self):
+        err = self._lib.norddrop_destroy(self._instance)
+        if err != 0:
+            err_type = LibResult(err).name
+            raise Exception(f"norddrop_destory has failed with code: {err}({err_type})")
+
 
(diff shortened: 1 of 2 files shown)


LukasPukenis:
Could we have a description in commit and PR why this fixes the segfault?

matszczygiel:
@LukasPukenis it's not finished yet, just wanted to see if the segfault still happens with these changes on the CI. I will update the description before marking it as ready

matszczygiel:
@LukasPukenis it's ready now :)

LukasPukenis (approved):
Epic! the description and code makes sense :)

PR #154, opened 2023-08-02
Lukasp/lock database test

LukasPukenis:


test/scenarios.py
```diff
@@ -6983,4 +6968,30 @@
...
+                    action.Start("172.20.0.5"),
+                    action.Repeated(
+                        [
+                            action.Parallel(
+                                [
+                                    action.NewTransfer(
+                                        "172.20.0.15", ["/tmp/testfile-small"]
+                                    ),
+                                    action.PurgeTransfersUntil(int(time.time() + 10)),
+                                    action.AssertTransfers([]),
```
matszczygiel, line 6986: The `AssertTransfers` should be executed at the very end. In here it can catch the transfer not being deleted from the other parallel run. And it actually showed up in the CI run

matszczygiel (approved):

test/scenarios.py
LukasPukenis: You're definitely right!

matszczygiel (approved):

PR #155, opened 2023-08-03
Race condition in database

matszczygiel:


drop-transfer/src/storage_dispatch.rs
```diff
@@ -295,8 +269,7 @@ impl From<&crate::Event> for Option<Event> {
                 file_id: file_id.to_string(),
                 by_peer: *by_peer,
             },
-            crate::Event::FileUploadPaused { .. } => return None,
-            crate::Event::FileDownloadPaused { .. } => return None,
+            _ => return None,
```
LukasPukenis, line 272: IMO wildcard matcher might bite back if we would implement a new event, the compiler wouldn't point to this place that it needs to be handled

drop-storage/src/lib.rs
```diff
@@ -488,8 +488,11 @@ impl Storage {
         let task = async {
             let conn = self.conn.lock().await;
             conn.execute(
-                "INSERT INTO outgoing_path_pending_states (path_id) VALUES ((SELECT id FROM \
-                 outgoing_paths WHERE transfer_id = ?1 AND path_hash = ?2))",
+                r#"
+                INSERT INTO outgoing_path_pending_states (path_id)
+                SELECT id
+                FROM outgoing_paths WHERE transfer_id = ?1 AND path_hash = ?2
+                "#,
```
LukasPukenis: Technically these queries might insert 0..N values, maybe it's worth asserting that we're expecting just 0 or 1?

LukasPukenis:
Slightly offtopic: we could have an additional test involving library stoppage as Android reported having issues with those

drop-storage/src/lib.rs
matszczygiel: Do you mean LIMIT 1 ?

matszczygiel:
Actually, tests were failing without stop() being fixed so I consider this to be covered

drop-storage/src/lib.rs
LukasPukenis: That's a good suggestion as well. I was concerned that we expect tat it should always be 1 item, then we found a race that it might be 0 which showed to us that there's an actual race condition in the codebase. Doing `LIMIT 1` would hide the fact that due to some mistake there are more than 1 entry.

LukasPukenis:
I still think it's worth having even if redundant tests. Android showed us that retry logic broke when the library was started and restarted, might as well just cover that specifically

drop-storage/src/lib.rs
matszczygiel: How do you want to achieve this?

matszczygiel:
We already have at least a couple of tests exercising that, scenario29-x

matszczygiel:
@LukasPukenis do you have any idea why 34 is failing? I cannot reproduce it on my local machine

LukasPukenis:
@matszczygiel judging from the logs the delete operation was not successful as it still left the transfer in there after returning from the API call, however we have a lock on our API layer so it probably means that our async task has not yet finished deleting the transfer data before returning. If that's the case then the testcase is broken

matszczygiel:
@LukasPukenis I think I fixed it on my last commit. The time + 10 since the test run on CI took more than 10 seconds and on my local machine less. Maybe we should use something like 0xffffffffffffffff ?

LukasPukenis (dismissed):
To me it looks a good improvement, I suggest @Lipt0nas to also take a look

Lipt0nas (approved):
Nice to see a little less of a reliance on events as well

drop-storage/src/lib.rs
matszczygiel: I don't know how to handle this. Leaving as it is

PR #156, opened 2023-08-07
Android team reported that the transfer can cancel itself

matszczygiel:
This changes how the cancel call is propagated. Instead of implicit API channel close we explicitly use the `Close` request

### DIFF
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -8,7 +8,7 @@ use std::{
 use anyhow::Context;
 use drop_config::DropConfig;
 use drop_storage::{sync, Storage};
-use slog::{error, info, warn, Logger};
+use slog::{debug, error, info, warn, Logger};
 use tokio::sync::{
     mpsc::{self, UnboundedSender},
     Mutex,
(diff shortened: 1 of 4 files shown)


matszczygiel:
The Android team is still testing but I guess it's fine to review

drop-transfer/src/manager.rs
```diff
@@ -588,7 +594,11 @@ impl TransferManager {
             )
             .await;
         state.xfer_sync.local = sync::TransferState::Canceled;
-        let _ = state.conn.take();
+
+        if let Some(conn) = state.conn.take() {
+            let _ = conn.send(ServerReq::Close);
+        }
```
LukasPukenis, line 600: What happens if `take()` returns `None` and how can it happen?

matszczygiel: It can happen when the peer is offline. Still, it's registered in the DB and is sent upon transfer resume

LukasPukenis (commented):
I have several questions regarding functionality:
- Does `Close` message introduction break the backwards compatibility?
- You mentioned that testing showed that not always cancel is issued by cancel API, can we craft such a test to ensure that this really fixes it?

matszczygiel:
The `Close` command is internal. Issued from API to the connection handler. It's 100% backward compatible since it's not affecting the wire behavior.
The thing is that we have appropriate tests. It's just not reproducible on our setup and on my local machine. Only the Android team managed to observe the bug

matszczygiel:
Upon inspecting the logs provided by the Android team the issue is gone

LukasPukenis (approved):

PR #157, opened 2023-08-07
[maybe done] Update moose tracker to v0.5.1

Lipt0nas:
Notable changes include the removal of the `Phase` field in all events, the addition of `TransferDirection` and `protocol_version` fields

### DIFF
--- a/drop-analytics/src/lib.rs
+++ b/drop-analytics/src/lib.rs
@@ -19,11 +19,11 @@ const MOOSE_STATUS_SUCCESS: i32 = 0;
 const MOOSE_VALUE_NONE: i32 = -1;
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
-pub enum Phase {
-    #[serde(rename = "start")]
-    Start,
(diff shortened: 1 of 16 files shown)


drop-analytics/src/lib.rs
```diff
@@ -43,14 +43,19 @@ pub struct FileInfo {
...
+        info: TransferInfo,
+        protocol_version: i32,
+    );
     fn service_quality_transfer_file(
         &self,
         res: Result<(), i32>,
-        phase: Phase,
         transfer_id: String,
         transfer_time: i32,
+        direction: TransferDirection,
```
matszczygiel, line 58: Changing API must have been quite a pain. If I were you I would simply ignore the argument inside while leaving the API untouched.
But since you already did this job it's great :+1:

matszczygiel:
Looks good :+1: 
The only problem I have is that it will cause massive merge conflicts since the dev left the main behind quite a bit

matszczygiel (approved):

PR #158, opened 2023-08-07
Some client connection errors should not abort the retires

matszczygiel:
Softening the transfer failure conditions

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -270,6 +270,17 @@ class Sleep(Action):
         return f"Sleep({self._seconds})"
 
 
+class SleepMs(Action):
+    def __init__(self, ms: int):
+        self._ms: int = ms
+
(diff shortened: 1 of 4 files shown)


test/drop_test/action.py
```diff
@@ -343,6 +354,25 @@ def __str__(self):
         return f"NoEvent({self._duration})"
 
 
+class AssertNoEventOfType(Action):
+    def __init__(self, forbidded: typing.List[type], duration: int = 6):
```
Lipt0nas, line 358:
```suggestion
    def __init__(self, forbidden: typing.List[type], duration: int = 6):
```
Small typo

Lipt0nas (approved):
nice :+1:

test/drop_test/action.py
matszczygiel: fixed

PR #159, opened 2023-08-07
Lukasp/testsuite improvements

LukasPukenis:
Testsuite has now the ability to catch libdrop's API errors. Also fixed testrunner locking issue

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -69,6 +69,25 @@ class Action:
         raise NotImplementedError("run() on base Action class")
 
 
+class ExpectError(Action):
+    def __init__(self, action: Action, err: int):
+        self._action = action
+        self._err = err
+
+    async def run(self, drop: ffi.Drop):
+        try:
+            await self._action.run(drop)
+        except ffi.DropException as e:
+            if self._err != e.error():
+                raise Exception(
+                    f"Received DropException with error: {self._err}, expected {e.error()} instead"
+                )
+            else:
+                return
+
+        raise Exception(f"Action must have thrown DropException")
+
+
 class Parallel(Action):
     def __init__(self, actions: typing.List[Action]):
         self._actions = actions
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -114,12 +114,9 @@ class NewTransfer(Action):
         self._paths: list[str] = paths
 
     async def run(self, drop: ffi.Drop):
-        UUIDS_LOCK.acquire()
-
-        xfid = drop.new_transfer(self._peer, self._paths)
-        UUIDS.append(xfid)
-
-        UUIDS_LOCK.release()
+        with UUIDS_LOCK:
+            xfid = drop.new_transfer(self._peer, self._paths)
+            UUIDS.append(xfid)
 
     def __str__(self):
         return f"NewTransfer({self._peer}, {self._paths})"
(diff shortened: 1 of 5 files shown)


test/drop_test/ffi.py
```diff
@@ -480,27 +497,33 @@ def start(self, addr: str, dbpath: str, max_reqs_per_sec: int):
...
         return ctypes.string_at(version).decode("utf-8")
 
     def __del__(self):
         err = self._lib.norddrop_destroy(self._instance)
         if err != 0:
             err_type = LibResult(err).name
-            raise Exception(f"norddrop_destory has failed with code: {err}({err_type})")
+            raise DropException(
+                f"norddrop_destory has failed with code: {err}({err_type})", err
+            )
```
matszczygiel, line 526: Is it OK to throw in the destructor in Python?

test/scenarios.py
```diff
@@ -6766,6 +6766,132 @@
             ),
         },
     ),
+    Scenario(
+        "scenario31-3",
+        "Remove file on sending side without rejecting it first. Expect error and file present in JSON output",
```
matszczygiel, line 6771: Can you clarify the description here that it's about removing the file from the DB, not physically from the disk?

test/drop_test/ffi.py
LukasPukenis: Honestly, no idea didn't even think about it. Quick googling doesn't reveal anything as well. Do you have any idea about possible side effects?

test/drop_test/action.py
```diff
@@ -545,9 +552,8 @@ def __init__(self, uuid_slot: int, fid):
         self._fid = fid
 
     async def run(self, drop: ffi.Drop):
-        UUIDS_LOCK.acquire()
-        drop.remove_transfer_file(UUIDS[self._uuid_slot], self._fid)
-        UUIDS_LOCK.release()
+        with UUIDS_LOCK:
```
matszczygiel, line 555: Me like it :+1:

matszczygiel (approved):
Nice improvement :+1:

test/drop_test/action.py
LukasPukenis: That was actually the thing that was blocking me when I mentioned it. A ton of `print()` later I found it :D

test/drop_test/ffi.py
matszczygiel: No, but I'm not that deep into Python. In Rust or C++ it's a severe offense

test/scenarios.py
LukasPukenis: Sure! I've clarified that it's about removal from the transfer itself

test/drop_test/ffi.py
LukasPukenis: I also have no idea about python, can't really find anything. I thought about maybe doing the creation and destructor explicitly but that doesn't seem worth it to me as that's only done once in the app lifecycle

matszczygiel: Yeah, so let's leave it as is. We will continue the discussion when it breaks in any way

matszczygiel (dismissed):

test/scenarios.py
```diff
@@ -6766,6 +6766,257 @@
...
+                            {
+                                event.File(
+                                    FILES["testfile-small"].id,
+                                    "testfile-small",
+                                    1048576,
+                                ),
+                            },
+                        )
+                    ),
+                    action.Sleep(2),
```
matszczygiel, line 6789: When do we need to sleep here?

LukasPukenis: Removed

matszczygiel (approved):

PR #161, opened 2023-08-07
Identify and fix causes of a very long time it takes to finish stop() call

matszczygiel:
Main changes:
* Use `tokio::JoinSet<()>` for efficient tasks cancelation
* 'Asyncify' checksum computation
* Ensure the `stop()` timing with two test cases

### DIFF
--- a/drop-transfer/src/ws/server/v4.rs
+++ b/drop-transfer/src/ws/server/v4.rs
@@ -515,7 +515,7 @@ impl handler::Downloader for Downloader {
         );
 
         // Check if we can resume the temporary file
-        match tokio::task::block_in_place(|| super::TmpFileState::load(&tmp_location.0)) {
+        match super::TmpFileState::load(&tmp_location.0).await {
             Ok(super::TmpFileState { meta, csum }) => {
                 debug!(
                     self.logger,
@@ -616,11 +616,8 @@ impl handler::Downloader for Downloader {
     }
 
     async fn validate(&mut self, path: &Hidden<PathBuf>) -> crate::Result<()> {
-        let csum = tokio::task::block_in_place(|| {
-            let file = std::fs::File::open(&path.0)?;
-            let csum = file::checksum(&mut io::BufReader::new(file))?;
-            crate::Result::Ok(csum)
-        })?;
+        let file = std::fs::File::open(&path.0)?;
+        let csum = file::checksum(&mut io::BufReader::new(file)).await?;
 
         if self.full_csum.get().await != csum {
             return Err(crate::Error::ChecksumMismatch);
--- a/drop-transfer/src/ws/server/v4.rs
+++ b/drop-transfer/src/ws/server/v4.rs
@@ -10,7 +10,7 @@ use futures::{SinkExt, StreamExt};
 use slog::{debug, error, info, warn};
 use tokio::{
     sync::mpsc::{self, Sender, UnboundedSender},
-    task::JoinHandle,
+    task::{AbortHandle, JoinSet},
 };
 use warp::ws::{Message, WebSocket};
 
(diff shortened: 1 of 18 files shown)


test/scenarios.py
```diff
@@ -2171,6 +2171,392 @@
             ),
         },
     ),
+    Scenario(
+        "scenario11-2",
+        "Initate a buch of transfer and stop the sender. Expect immediate cancelation",
```
LukasPukenis, line 2176: Maybe we could use `action.Repeated` for creating N transfers with the same file, then awaiting for at least several of them to start(now its done using count=4) and then stop? This would reduce the testcode and would make it more obvious what is being tested.

I would also add this to the description that we're deliberately waiting for some
transfers to start so it would be more obvious instead of scanning through the scenario code :)

LukasPukenis, line 2176: Maybe we could use `action.Repeated` for creating N transfers with the same file, then awaiting for at least several of them to start(now its done using count=4) and then stop? This would reduce the testcode and would make it more obvious what is being tested.

I would also add this to the description that we're deliberately waiting for some
transfers to start so it would be more obvious instead of scanning through the scenario code :)

drop-transfer/src/ws/server/v4.rs
```diff
@@ -616,11 +616,8 @@ impl handler::Downloader for Downloader {
     }
 
     async fn validate(&mut self, path: &Hidden<PathBuf>) -> crate::Result<()> {
-        let csum = tokio::task::block_in_place(|| {
-            let file = std::fs::File::open(&path.0)?;
-            let csum = file::checksum(&mut io::BufReader::new(file))?;
-            crate::Result::Ok(csum)
-        })?;
+        let file = std::fs::File::open(&path.0)?;
+        let csum = file::checksum(&mut io::BufReader::new(file)).await?;
```
LukasPukenis, line 650: Just an idea, I am not familiar with tokio that well - is it possible to drive the task manually without writing our own executor? This would be nice to check that the `checksum` is actually behaving properly.

It's not necessary, just a thought

LukasPukenis (commented):
First two commits(tests and fix) are totally legit. could you split the third commit into separate for a better review?

drop-transfer/src/ws/server/v4.rs
matszczygiel: I don't know what you mean exactly but you can call https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll on the future

LukasPukenis: Looks exactly what I wanted :) I was just curious if we can test that `checksum()` requires several invocations in order to complete. Because we know that the internal buffer is 8K if we would pass in a 20K file it should take exactly three invocations.

That would be nice to have a unit test for

test/drop_test/action.py
```diff
@@ -618,3 +618,24 @@ async def run(self, drop: ffi.Drop):
...
+    async def run(self, drop: ffi.Drop):
+        start = time.time()
+        await self._action.run(drop)
+        elapsed = time.time() - start
+
+        expected = float(self._ms) / float(1000)
+
+        if elapsed > expected:
+            raise Exception(
+                f"Action took longer ({elapsed} s) than expected ({expected} s)"
```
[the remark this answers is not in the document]
matszczygiel: dome

test/scenarios.py

drop-transfer/src/ws/server/v4.rs
matszczygiel: Done. I must admit it was quite a fun

LukasPukenis:
@matszczygiel the test and fix are legit, but the commit - 6a5f2e7ae281c9e9f3d497bc97ced6b446e87bf5 is a refactor I assume? It could be moved to a different PR as now I am sure about the fix and test and those could be easily merged, but the refactoring commit is so huge that it's a separate effort to review :)

matszczygiel:
@LukasPukenis it's not a refactor. The `JoinSet` (almost) guarantees the cancelation of all tasks spawned. That was not the case before. It contributes to the fix

LukasPukenis (approved):

PR #162, opened 2023-08-09
Update Moose to v0.5.1

Lipt0nas:
Notable changes include the removal of the `Phase` field in all events, the addition of `TransferDirection` and `protocol_version` fields

In addtion, we had to downgrade our version of `chrono` to satisfy the version requested by moose

### DIFF
--- a/drop-analytics/src/lib.rs
+++ b/drop-analytics/src/lib.rs
@@ -19,11 +19,11 @@ const MOOSE_STATUS_SUCCESS: i32 = 0;
 const MOOSE_VALUE_NONE: i32 = -1;
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize)]
-pub enum Phase {
-    #[serde(rename = "start")]
-    Start,
-    #[serde(rename = "end")]
-    End,
+pub enum TransferDirection {
+    #[serde(rename = "upload")]
+    Upload,
+    #[serde(rename = "download")]
+    Download,
 }
 
 #[derive(Serialize, Deserialize)]
@@ -43,14 +43,19 @@ pub struct FileInfo {
 }
 
 pub trait Moose: Send + Sync {
-    fn service_quality_initialization_init(&self, res: Result<(), i32>, phase: Phase);
-    fn service_quality_transfer_batch(&self, phase: Phase, transfer_id: String, info: TransferInfo);
+    fn service_quality_initialization_init(&self, res: Result<(), i32>);
+    fn service_quality_transfer_batch(
+        &self,
+        transfer_id: String,
+        info: TransferInfo,
+        protocol_version: i32,
+    );
     fn service_quality_transfer_file(
         &self,
         res: Result<(), i32>,
-        phase: Phase,
         transfer_id: String,
         transfer_time: i32,
+        direction: TransferDirection,
         info: Option<FileInfo>,
     );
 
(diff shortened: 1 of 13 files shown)


drop-transfer/src/ws/client/mod.rs
```diff
@@ -92,6 +92,10 @@ async fn connect_to_peer(
         }
     };
 
+    state
+        .moose
+        .service_quality_transfer_batch(xfer.id().to_string(), xfer.info(), ver.into());
+
```
LukasPukenis, line 98: This was moved from another place, could you explain the reasoning?

Lipt0nas: this is the earliest point, in which we can get the protocol version, to provide to the event

drop-analytics/src/lib.rs
```diff
@@ -19,11 +19,11 @@ const MOOSE_STATUS_SUCCESS: i32 = 0;
...
-pub enum Phase {
-    #[serde(rename = "start")]
-    Start,
-    #[serde(rename = "end")]
-    End,
+pub enum TransferDirection {
+    #[serde(rename = "upload")]
+    Upload,
+    #[serde(rename = "download")]
+    Download,
```
LukasPukenis, line 26: We have 4 different combinations: upload can start/stop, and the same for download. However we only had two(start, end), now we have two as well but different(upload, download). My concern is that we effectively lose some information that might be desired by analytics team. If we send two events(on start and end) then how do analytics differentiate between the two?

drop-transfer/src/ws/client/mod.rs
LukasPukenis: Oh okay, makes sense. I guess we report the version that's used for the transaction, meaning if we maintain backwards compatibility and we have v5 and v1 communicating, it's gonna report v1? Just checking as it makes sense.

Is there a possibility to report the highest supported version as well so in this case it would report both 5 and 1?

drop-transfer/src/file/mod.rs
```diff
@@ -31,6 +31,7 @@ pub trait File {
     fn subpath(&self) -> &FileSubPath;
     fn size(&self) -> u64;
     fn mime_type(&self) -> &str;
+    fn direction(&self) -> TransferDirection;
```
matszczygiel, line 34: Using `self` here requires you to do a lookup of the file. This is not necessary. I suggest removing `self` from here or even better make it a `Transfer` trait method

drop-transfer/src/ws/client/mod.rs
Lipt0nas: Yes, in the case of v5 and v1 communicating, it would report v1, including highest supported version would require another revision of the tracker, but we also include the lib version in the data, so we could infer that from it

LukasPukenis: That's right!

drop-transfer/src/file/mod.rs
Lipt0nas: changed it to not take `Self`, making it part of the transfer trait would have been a little more tricky, since as far as i know, there is no way to partially specialize a trait

Lipt0nas: I think i understood the initial comment incorrectly, fixed now :+1:

matszczygiel (dismissed):
I see that closing the moose file calls in the file event emitter was a good move. It probably helped in the preparation of this PR.
LGTM :+1:

LukasPukenis (approved):
Looks nice, @matszczygiel also gave thumbs up, thus merging

PR #163, opened 2023-08-09
Allow constructing outgoing files without the access to the file system.

matszczygiel:
This allows for a fairly easy reconstruction of transfer from the database upon instance startup. Previously when restoring a file failed the whole transfer was ignored since it wasn't possible to restore its data structure. With these changes, the structure is restored and the file system is only consulted during the regular operation of the connection handler.

Additionally, I've found a bug with failed and started events sometimes being not emitted on the server side

### DIFF
--- a/drop-transfer/src/ws/server/v2.rs
+++ b/drop-transfer/src/ws/server/v2.rs
@@ -192,7 +192,7 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
                     let file_id = self
                         .xfer
                         .file_by_subpath(&file)
-                        .expect("File should be there sice we have a task registered")
+                        .expect("File should be there since we have a task registered")
                         .id();
 
                     debug!(
@@ -206,7 +206,7 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
                     if let Err(err) = self
                         .state
                         .transfer_manager
-                        .incoming_finsh_download(self.xfer.id(), file_id)
+                        .incoming_finish_download(self.xfer.id(), file_id)
                         .await
                     {
                         warn!(self.logger, "Failed to store download finish: {err}");
--- a/drop-transfer/src/ws/server/v2.rs
+++ b/drop-transfer/src/ws/server/v2.rs
@@ -189,7 +189,28 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
             }) = self.jobs.remove(&file)
             {
                 if !task.is_finished() {
+                    let file_id = self
+                        .xfer
+                        .file_by_subpath(&file)
+                        .expect("File should be there sice we have a task registered")
+                        .id();
+
+                    debug!(
+                        self.logger,
+                        "Aborting download job: {}:{file_id}",
+                        self.xfer.id()
+                    );
+
                     task.abort();
+
+                    if let Err(err) = self
+                        .state
+                        .transfer_manager
+                        .incoming_finsh_download(self.xfer.id(), file_id)
+                        .await
+                    {
+                        warn!(self.logger, "Failed to store download finish: {err}");
+                    }
 
                     events
                         .failed(crate::Error::BadTransferState(format!(
--- a/test/run.py
+++ b/test/run.py
@@ -67,7 +67,7 @@ def cleanup_files(files: typing.Dict):
         ) as proc:
             proc.communicate()
             if proc.returncode != 0:
-                raise Exception("Couldn't create a testfile")
+                print(f"Couldn't remove a testfile {name}")
 
 
 async def main():
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -6758,7 +6758,7 @@ scenarios = [
                     action.Wait(event.Progress(0, FILES["testfile-big"].id)),
                     action.Stop(),
                     action.DeleteFileFromFS("/tmp/testfile-big"),
-                    # restrat
+                    # restart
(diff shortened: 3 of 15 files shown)


test/run.py
```diff
@@ -67,7 +67,7 @@ def cleanup_files(files: typing.Dict):
         ) as proc:
             proc.communicate()
             if proc.returncode != 0:
-                raise Exception("Couldn't create a testfile")
+                print(f"Couldn't remove a testfile {name}")
```
LukasPukenis, line 70: Was this hurting in some way? An exception is nice because else the test would fail as a consequence

test/scenarios.py
```diff
@@ -6730,6 +6730,84 @@
             ),
         },
     ),
+    Scenario(
+        "scenario29-13",
```
LukasPukenis, line 6734: This is a good test. It also highlights that:
- the sender is oblivious that a file was removed from the filesystem
- probably no failed state is inserted into the database on the sender side

drop-transfer/src/ws/server/v2.rs
```diff
@@ -189,8 +189,29 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
             }) = self.jobs.remove(&file)
             {
                 if !task.is_finished() {
+                    let file_id = self
+                        .xfer
+                        .file_by_subpath(&file)
+                        .expect("File should be there sice we have a task registered")
+                        .id();
```
LukasPukenis, line 196:
```suggestion
                        .file_by_subpath(&file)
                        .expect("File should be there since we have a task registered")
                        .id();
```

test/scenarios.py
```diff
@@ -6730,6 +6730,84 @@
...
+                    ),
+                    action.Wait(event.Start(0, FILES["testfile-big"].id)),
+                    # wait for the initial progress indicating that we start from the beginning
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id, 0)),
+                    # make sure we have received something, so that we have non-empty tmp file
+                    action.Wait(event.Progress(0, FILES["testfile-big"].id)),
+                    action.Stop(),
+                    action.DeleteFileFromFS("/tmp/testfile-big"),
+                    # restrat
+                    action.Start("172.20.0.5", dbpath="/tmp/db/29-13-ren.sqlite"),
```
LukasPukenis, line 6762:
```suggestion
                    action.DeleteFileFromFS("/tmp/testfile-big"),
                    # restart
                    action.Start("172.20.0.5", dbpath="/tmp/db/29-13-ren.sqlite"),
```

drop-transfer/src/ws/server/v2.rs
```diff
@@ -189,8 +189,29 @@ impl<const PING: bool> HandlerLoop<'_, PING> {
...
+                        self.xfer.id()
+                    );
+
                     task.abort();
 
+                    if let Err(err) = self
+                        .state
+                        .transfer_manager
+                        .incoming_finsh_download(self.xfer.id(), file_id)
+                        .await
```
LukasPukenis, line 210: the method is misspelled, it should be `incoming_finish_download()`

LukasPukenis (approved):
Looks like a nice simplification of the File handling though I left a comment on testcase where it seems that the sender has no new state in case of file failure while it should

test/run.py
matszczygiel: Yeah. I wrote a test with one of the files being deleted as part of the test. This was clashing with my test

test/scenarios.py
matszczygiel: Yeah, I'm not sure how to fix that or is it worth fixing

drop-transfer/src/ws/server/v2.rs
matszczygiel: done

test/scenarios.py

drop-transfer/src/ws/server/v2.rs
matszczygiel: done

test/run.py
LukasPukenis: But it should really ever error if the creation has failed like for example the testfile path is wrong, permission issues or etc, how does it get involved with file deletion?

test/scenarios.py
LukasPukenis: I've approved the PR still as I agree that this might be minor(who really deletes the files when sending), however, it shouldn't be too hard to go via failed file transfer way?

matszczygiel: It actually is.
So the sending of chunks is initiated by the receiver and if it finds a temporary it requests for the checksum.
The checksum fails on the sender and is communicated back to the receiver.
The receiver emits an error event and what's crucial **does not issue a download message to the sender**.

So in the end the receiver knows that the file transfer failed but the sender doesn't because it really failed before he was made aware of the upload at all.

As I mentioned it should behave differently but I honestly don't know how involving fixing it would be. Probably would require introducing a new protocol version so that the sender can relate checksum requests with a particular transfer which we will do eventually anyway.
So I would leave it as a ticket for 5.1

LukasPukenis: I think this is a separate issue. I meant that if the sender initiates the transfer, goes offline, deletes the file and comes online he should expect FileFailed state in his database

matszczygiel: I mean it's basically the same because we insert state into DB based on events emitted

PR #167, opened 2023-08-17
Implement file soft deletion, allow deleting files upon reaching fail…

matszczygiel:
…ed, rejected or completed state.

### DIFF
--- a/drop-storage/migrations/003-retries/up.sql
+++ b/drop-storage/migrations/003-retries/up.sql
@@ -43,3 +43,7 @@ CREATE TABLE IF NOT EXISTS sync_incoming_files_inflight (
   FOREIGN KEY(sync_id, path_id) REFERENCES sync_incoming_files(sync_id, path_id) ON DELETE CASCADE ON UPDATE CASCADE
 );
 
+
+ -- paths soft deletion
+ALTER TABLE incoming_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT FALSE CHECK (is_deleted IN (FALSE, TRUE));
+ALTER TABLE outgoing_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT FALSE CHECK (is_deleted IN (FALSE, TRUE));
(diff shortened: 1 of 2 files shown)


drop-storage/migrations/003-retries/up.sql
```diff
@@ -43,3 +43,7 @@ CREATE TABLE IF NOT EXISTS sync_incoming_files_inflight (
   FOREIGN KEY(sync_id, path_id) REFERENCES sync_incoming_files(sync_id, path_id) ON DELETE CASCADE ON UPDATE CASCADE
 );
 
+
+ -- paths soft deletion
+ALTER TABLE incoming_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT FALSE;
+ALTER TABLE outgoing_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT FALSE;
```
LukasPukenis, line 49:
```suggestion
ALTER TABLE incoming_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT 0 CHECK (is_deleted IN (0, 1));
ALTER TABLE outgoing_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT 0 CHECK (is_deleted IN (0, 1));
```

LukasPukenis (dismissed):
Looks like a solid PR

drop-storage/migrations/003-retries/up.sql
matszczygiel: Do you see any advantage of using `0 and 1` over `FALSE and TRUE`? It states intent better IMHO

LukasPukenis: I totally agree, it's just making sure that accidentally we wouldn't put in a non [0;1] number in there in some query. Maybe doing `is_deleted IN (TRUE, FALSE)` would work better but that's a speculation from my end :) You can freely dismiss this comment

matszczygiel: Done

PR #169, opened 2023-08-18
Ensure `Paused` events are emited when stop() is called

matszczygiel:
Emitting events in the handlers `Drop` implementation ensures they are constantly emitted.
This is a preparation for storing progress in the events, so test fixes and including the file progress in the events are going to be included in subsequent PRs


LukasPukenis (commented):
This could benefit from a testcase

matszczygiel:
@LukasPukenis we have plenty of tests for `stop()` calls, and these are failing because of this change (no events are expected in the tests yet it is emitted).
I'm going to fix those in the next PR since the structure of events will also change rendering the test cases invalid again.
So in essence I don't want to fix the tests twice, so I'm leaving it for the very end.

LukasPukenis:
@matszczygiel thanks for clarifying, it's all good then

LukasPukenis (approved):

PR #171, opened 2023-08-21
Store bytes transfered with paused and reject states

matszczygiel:
@LukasPukenis this is the continuation of progress recovery.

I'm not sure about some things:
* is it better to provide global per file field "bytes_transfered" in the JSON API or leave it as it is, as progress inside each state change
* the same question but with the database itself, should we store progress globally or in the `*_states` tables

IMHO it might be beneficial to store and present in the JSON progress as a global per-file thing

### DIFF
--- a/drop-storage/migrations/003-retries/up.sql
+++ b/drop-storage/migrations/003-retries/up.sql
@@ -47,3 +47,24 @@ CREATE TABLE IF NOT EXISTS sync_incoming_files_inflight (
  -- paths soft deletion
 ALTER TABLE incoming_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT FALSE CHECK (is_deleted IN (FALSE, TRUE));
 ALTER TABLE outgoing_paths ADD COLUMN is_deleted INTEGER NOT NULL DEFAULT FALSE CHECK (is_deleted IN (FALSE, TRUE));
+
+
+ -- storing progress
+ALTER TABLE incoming_path_reject_states ADD COLUMN bytes_received INTEGER NOT NULL DEFAULT 0;
+ALTER TABLE outgoing_path_reject_states ADD COLUMN bytes_sent     INTEGER NOT NULL DEFAULT 0;
+
+CREATE TABLE IF NOT EXISTS incoming_path_paused_states (
+  path_id INTEGER NOT NULL,
+  bytes_received INTEGER NOT NULL,
+  created_at TIMESTAMP NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+  FOREIGN KEY(path_id) REFERENCES incoming_paths(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS outgoing_path_paused_states (
+  path_id INTEGER NOT NULL,
+  bytes_sent INTEGER NOT NULL,
+  created_at TIMESTAMP NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+  FOREIGN KEY(path_id) REFERENCES outgoing_paths(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+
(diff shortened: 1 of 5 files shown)


drop-storage/migrations/003-retries/up.sql
```diff
@@ -47,3 +47,24 @@ CREATE TABLE IF NOT EXISTS sync_incoming_files_inflight (
...
+  FOREIGN KEY(path_id) REFERENCES incoming_paths(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE TABLE IF NOT EXISTS outgoing_path_paused_states (
+  path_id INTEGER NOT NULL,
+  bytes_sent INTEGER NOT NULL,
+  created_at TIMESTAMP NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+  FOREIGN KEY(path_id) REFERENCES outgoing_paths(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
```
LukasPukenis, line 69: `*_path_reject_states` contain `DEFAULT 0` but `*_path_paused_states` do not, is there a reason for the difference?

test/scenarios.py
```diff
@@ -3821,6 +3821,18 @@
...
+                                0,
+                                FILES[
+                                    "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456.txt"
+                                ].id,
+                            ),
+                            event.Start(
+                                1,
+                                FILES[
+                                    "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456.txt"
+                                ].id,
```
LukasPukenis, line 3834: Could you explain a bit more why this PR triggered a change in this testcase?

test/scenarios.py
```diff
@@ -5773,6 +5847,56 @@
                     action.CancelTransferRequest(0),
```
LukasPukenis, line 5847: The receiver should also receive Paused event if the sender did the `stop()` and vice versa

LukasPukenis (commented):
The tables look fine, so it should contain the changes we need. I am a bit worried about the testcases where the other side is not tested for Paused event

LukasPukenis:
@matszczygiel I think storing the progress in Paused tables is totally fine and preferred way, it stores all the information at checkpoints so is very simple IMO :)

drop-storage/migrations/003-retries/up.sql
matszczygiel: Yes, the bytes_* for rejected states is created by `ALTER TABLE` meaning if there is data in the table already which value should it contain? Without `DEFAULT 0` the migration script would fail

test/scenarios.py
matszczygiel: I'm not sure, probably the test was failing on dev already and we simply were ignoring this

drop-storage/migrations/003-retries/up.sql
LukasPukenis: Good point, makes sense

test/scenarios.py
matszczygiel: It does https://github.com/NordSecurity/libdrop/pull/171/files/614465fcdf351d7d06de0b936214015313ea4d73#diff-01281065a74eabdb084c366fb316dda3f5defaa50ed1f54df6bcac6209f35297L5759

matszczygiel:
@LukasPukenis what do you mean by preferred? I this it doesn't matter from the API point of view but would result in a simpler logic if we'd store it globally (but updated of course on checkpoints)

LukasPukenis:
@matszczygiel I think it's totally fine to just have paused events. If you fancy then you can calculate per-file field of bytes_received/bytes_sent for sure as well inside the final JSON, it might be easier for API consumers.

LukasPukenis (approved):

PR #178, opened 2023-08-24
Check and remove unused error codes

matszczygiel:
It turns out the `IncomingTransferFailed` event was not used anymore.
Surprisingly there was no need for a general `InternalError` status since all cases are pretty much well-covered


drop-core/src/status.rs
```diff
@@ -24,9 +19,8 @@ pub enum Status {
     DbLost = 32,
     FileChecksumMismatch = 33,
     FileRejected = 34,
-    FilePaused = 35,
-    FileFailed = 36,
-    FileFinished = 37,
+    FileFailed = 35,
+    FileFinished = 36,
```
LukasPukenis, line 23: How long these values have been here? Their numerical value has changed

test/drop_test/error.py
```diff
@@ -36,9 +20,8 @@ class Error(IntEnum):
     DB_LOST = (32,)
     FILE_CHECKSUM_MISMATCH = (33,)
     FILE_REJECTED = (34,)
-    FILE_PAUSED = (35,)
-    FILE_FAILED = (36,)
-    FILE_FINISHED = (37,)
+    FILE_FAILED = (35,)
+    FILE_FINISHED = (36,)
```
LukasPukenis, line 24: Changing numerical values of these breaks the API

matszczygiel: Yeah, I'm aware. These were introduced lately and never reached the main branch, thus I considered it safe to change

LukasPukenis: Okay. Please update v5 docs as well

PR #180, opened 2023-08-24
fix panic when stop is called immediately after download

matszczygiel:
This is fixed by locking the instance pointer within the libdrop method call and then passing it to the async task.
This avoids concurrent access to the lock, thus eliminating the possibility of reordering locking with respect to method calls

### DIFF
--- a/norddrop/src/device/mod.rs
+++ b/norddrop/src/device/mod.rs
@@ -160,6 +160,11 @@ impl NordDropFFI {
             }
         });
 
+        let mut instance = self.instance.clone().blocking_lock_owned();
+        if instance.is_some() {
+            return Err(ffi::types::NORDDROP_RES_INSTANCE_START);
+        }
+
         self.rt.block_on(async {
             let service = match Service::start(
                 addr,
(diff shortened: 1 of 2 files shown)


LukasPukenis:
I sort of understand what is happening but not 100% sure why the previous version could have panicked. Could you explain in more technical terms?

matszczygiel:
@LukasPukenis so the case was that calling `download()` and `stop()` right after sometimes caused panic `Instance not initialized`
The two facts are:
* instance is stored as more or less `Mutex<Option<Instance>>` so all calls must check if the instance is set or no
* `stop()` call basically sets the `Option` to `None`

So what happens is that calling `download()` does not fail cleanly but with a panic when the instance is not set, which is quite OK, since you're not allowed to call it without initializing the instance.
The thing is that `download()` first spawns an async task and then performs the check. So when you call `stop()` right after, it might finish setting the instance to `None` before the `download()` call manages to spawn the async task.
Ensuring the check is done before `download()` returns should fix the issue

LukasPukenis:
That's a nice catch, thanks for explaining

LukasPukenis (approved):

PR #183, opened 2023-08-28
Fix receiving peer cannot know after restarting which files where accepted

matszczygiel:
This PR uses `Pending` states as a way to inform the app about the file being accepted.

What is more, I've decided that we can store `base_dir` for downloads in the `Pending` states table in contrast to the `Started` states table, since it's more tied to the `download()` call. Details in the commit message


Lipt0nas (approved):
Really nice idea to also change where `base_dir` is being stored

PR #188, opened 2023-08-28
More tests regarding FFI errors and resume behavior

LukasPukenis:
This PR swaps around `manager::resume` and WebSocket server start. This prevents from broken `norddrop_start()` from trying to communicate with the peer already, failing to accept incoming connections, and producing errors

### DIFF
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -960,12 +960,25 @@ pub(crate) async fn resume(
     guard: &AliveGuard,
     stop: &CancellationToken,
 ) {
-    state.storage.cleanup_garbage_transfers().await;
+    let xfers = state.transfer_manager.outgoing.lock().await;
 
-    *state.transfer_manager.incoming.lock().await =
-        restore_incoming(&state.storage, &state.config, logger).await;
-    *state.transfer_manager.outgoing.lock().await =
-        restore_outgoing(state, logger, guard, stop).await;
+    for xstate in xfers.values() {
+        ws::client::spawn(
+            state.clone(),
+            xstate.xfer.clone(),
+            logger.clone(),
+            guard.clone(),
+            stop.clone(),
+        );
+    }
+}
+
+pub(crate) async fn restore_transfers_state(state: &Arc<State>, logger: &Logger) {
+    let incoming = restore_incoming(&state.storage, &state.config, logger).await;
+    *state.transfer_manager.incoming.lock().await = incoming;
+
+    let outgoing = restore_outgoing(state, logger).await;
+    *state.transfer_manager.outgoing.lock().await = outgoing;
 }
 
 async fn restore_incoming(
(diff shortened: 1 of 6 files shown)


drop-transfer/src/service.rs
```diff
@@ -77,16 +77,16 @@ impl Service {
...
                 addr,
                 state.clone(),
                 auth,
                 logger.clone(),
                 stop.clone(),
-                guard,
+                guard.clone(),
             )?;
 
+            manager::resume(&state, &logger, &guard, &stop).await;
```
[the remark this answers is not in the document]
LukasPukenis: I actually remember but I don't know the specifics, could you write those down?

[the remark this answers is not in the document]
LukasPukenis: @matszczygiel I've updated accordingly

drop-transfer/src/manager.rs
```diff
@@ -960,10 +960,30 @@ pub(crate) async fn resume(
...
+pub(crate) async fn restore_transfer_state(
+    state: &Arc<State>,
+    logger: &Logger,
+    guard: &AliveGuard,
+    stop: &CancellationToken,
+) {
+    let incoming = restore_incoming(&state.storage, &state.config, logger).await;
+    *state.transfer_manager.incoming.lock().await = incoming;
+
+    let outgoing = restore_outgoing(state, logger, guard, stop).await;
```
matszczygiel, line 985: I bet `stop` and `guard` are unused variables in this function

drop-transfer/src/manager.rs
```diff
@@ -960,10 +960,30 @@ pub(crate) async fn resume(
...
+            state.clone(),
+            xstate.xfer.clone(),
+            logger.clone(),
+            guard.clone(),
+            stop.clone(),
+        );
+    }
+}
+
+pub(crate) async fn restore_transfer_state(
```
matszczygiel, line 976:
```suggestion
pub(crate) async fn restore_transfers_state(
```

LukasPukenis: +1

drop-transfer/src/manager.rs
```diff
@@ -961,11 +961,31 @@ pub(crate) async fn resume(
     stop: &CancellationToken,
 ) {
     state.storage.cleanup_garbage_transfers().await;
```
matszczygiel, line 963: Let's move it to the beginning of the `restore_transfers_state()` so that we do not restore transfers to be deleted

drop-transfer/src/manager.rs
```diff
@@ -961,11 +961,31 @@ pub(crate) async fn resume(
...
+            stop.clone(),
+        );
+    }
+}
+
+pub(crate) async fn restore_transfers_state(
+    state: &Arc<State>,
+    logger: &Logger,
+    guard: &AliveGuard,
+    stop: &CancellationToken,
```
matszczygiel, line 982:
```suggestion
```

drop-transfer/src/manager.rs
LukasPukenis: I've moved it outside into service::start() just before the restoration

matszczygiel (approved):

PR #189, opened 2023-08-29
Fix: Android reported a huge 17 min delay for receiving paused events

matszczygiel:
This PR abstracts away the socket into a custom type that imposes timeouts automatically

### DIFF
--- a/drop-transfer/src/ws/client/socket.rs
+++ b/drop-transfer/src/ws/client/socket.rs
@@ -1,0 +1,60 @@
+use std::{
+    io,
+    time::{Duration, Instant},
+};
+
+use futures_util::{SinkExt, StreamExt};
+use tokio::net::TcpStream;
+use tokio_tungstenite::{tungstenite::Message, WebSocketStream};
+
+pub type WsStream = WebSocketStream<TcpStream>;
+
+pub struct WebSocket {
+    stream: WsStream,
+    recv_last: Option<Instant>,
+    recv_timeout: Duration,
+    send_timeout: Duration,
+}
+
+impl WebSocket {
+    pub fn new(stream: WsStream, recv_timeout: Duration, send_timeout: Duration) -> Self {
+        Self {
+            stream,
+            recv_last: None,
+            recv_timeout,
+            send_timeout,
+        }
+    }
+
+    pub async fn send(&mut self, msg: Message) -> crate::Result<()> {
+        tokio::time::timeout(self.send_timeout, self.stream.send(msg))
+            .await
+            .map_err(|err| io::Error::new(io::ErrorKind::TimedOut, err))??;
+
+        Ok(())
+    }
+
+    pub async fn recv(&mut self) -> crate::Result<Message> {
+        let timeout = self.recv_last.map_or(self.recv_timeout, |last| {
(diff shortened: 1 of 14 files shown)


drop-transfer/src/ws/client/socket.rs
```diff
@@ -0,0 +1,56 @@
...
+    recv_last: Instant,
+    recv_timeout: Duration,
+    send_timeout: Duration,
+}
+
+impl WebSocket {
+    pub fn new(stream: WsStream, recv_timeout: Duration, send_timeout: Duration) -> Self {
+        Self {
+            stream,
+            recv_last: Instant::now(),
```
LukasPukenis, line 23: To be pedantic I think `Option<Instant>` would be slightly more correct

drop-transfer/src/ws/client/socket.rs
```diff
@@ -0,0 +1,56 @@
...
+            recv_timeout,
+            send_timeout,
+        }
+    }
+
+    pub async fn send(&mut self, msg: Message) -> crate::Result<()> {
+        tokio::time::timeout(self.send_timeout, self.stream.send(msg))
+            .await
+            .map_err(|err| io::Error::new(io::ErrorKind::TimedOut, err))??;
```
LukasPukenis, line 32: I am curious why do we need 2x`?`? Can `map_err` fail to convert the message..?

drop-transfer/src/ws/client/socket.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    pub async fn recv(&mut self) -> crate::Result<Message> {
+        let timeout = self.recv_timeout.saturating_sub(self.recv_last.elapsed());
+
+        let msg = tokio::time::timeout(timeout, self.stream.next())
+            .await
+            .map_err(|err| io::Error::new(io::ErrorKind::TimedOut, err))?
+            .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "Emptry socket stream"))??;
+
+        Ok(msg)
```
LukasPukenis, line 45: Similar question to above, if the `await` resulted in a timeout error we map it to `TimedOut` but why do we need another `ok_or_else` for?

LukasPukenis (commented):
Curious why those 17minute delays happened and how the existing timeout helps?

drop-transfer/src/ws/client/socket.rs
matszczygiel: Oh, yeah. True. You gave me a thought that I'm not setting this field in the receive() method. My mistake

drop-transfer/src/ws/client/socket.rs
matszczygiel: No, there are two levels of error here

LukasPukenis: Oh okay `stream.send` might error as well

drop-transfer/src/ws/client/socket.rs
matszczygiel: Fixed

matszczygiel:
@LukasPukenis the existing timeout was imposed on the `recv()` call. Yet Android managed to hang on the `send()` call

LukasPukenis (approved):

PR #191, opened 2023-08-30
Fix: removing transfer from the db prevents the state syncing since the data is lost

matszczygiel:
This one implements soft deletion for transfers.
The transfers are automatically hard deleted in the library initialization procedure once it's determined that the full state was synced with the peer


LukasPukenis (commented):
I would really like a test and especially an integration one. However, it can only be achieved by not hiding the transfer if it's marked as deleted, rightfully so.

Instead, at least we could have a unit test at least around `cleanup_garbage_transfer()`. What do you think?

matszczygiel:
Yeah, I will try to figure out something

LukasPukenis (dismissed):

matszczygiel:
@LukasPukenis I've added the unit test

LukasPukenis (approved):

PR #192, opened 2023-08-30
Purge transfers should check for state

matszczygiel:
Blocked by https://github.com/NordSecurity/libdrop/pull/191

### DIFF
--- a/drop-storage/src/lib.rs
+++ b/drop-storage/src/lib.rs
@@ -865,8 +865,14 @@ impl Storage {
         let task = async {
             let conn = self.conn.lock().await;
             conn.execute(
-                "UPDATE transfers SET is_deleted = TRUE WHERE created_at < datetime(?1, \
-                 'unixepoch')",
+                r#"
+                UPDATE transfers SET is_deleted = TRUE
+                WHERE created_at < datetime(?1, 'unixepoch')
+                    AND (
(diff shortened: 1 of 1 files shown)


drop-storage/src/lib.rs
```diff
@@ -1721,8 +1773,22 @@ mod tests {
             .purge_transfers(&[transfer_id_1.to_string(), transfer_id_2.to_string()])
             .await;
 
+        // Because the transfers havent' reached termial state
```
LukasPukenis, line 1776:
```suggestion
        // Because the transfers haven't reached the terminal state
```

LukasPukenis (approved):
These both PRs look good, however, they are tricky in their behavior. Would be nice to find a way to integrate-test them.

Few ideas:
- Emit internal events via API that would be useless to real API users but only for testing purposes(not that nice)
- Return deleted transfers in the JSON API and mark them as deleted. May return them in a separate object so as not to make API users accidentally use them(nicer)

What is your take on these?

matszczygiel:
@LukasPukenis in https://github.com/NordSecurity/libdrop/pull/191 I introduced a unit test to check how many transfers we've removed in the `remove_garbage_transfers()`. I think it somewhat checks all the possible cases

LukasPukenis (approved):

PR #194, opened 2023-08-31
Fix DDoS test case + improve when the protection is triggered

matszczygiel:
Warp filters were checked in order, so I reordered it so that the DDoS filter is triggered first

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -11,6 +11,7 @@ import json
 import time
 import glob
 import socket
+import requests
 
 from . import event, ffi
 from .logger import logger
@@ -102,6 +103,19 @@ class ExpectError(Action):
                 return
 
         raise Exception(f"Action must have thrown DropException")
+
+
+class ExpectAnyError(Action):
+    def __init__(self, action: Action):
+        self._action = action
+
+    async def run(self, drop: ffi.Drop):
+        try:
+            await self._action.run(drop)
+        except:
+            return
+
+        raise Exception("Action must have thrown an exception")
 
 
 class Parallel(Action):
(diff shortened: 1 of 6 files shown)


test/drop_test/action.py
```diff
@@ -104,6 +105,17 @@ async def run(self, drop: ffi.Drop):
...
 
+class ExpectAnyError(Action):
+    def __init__(self, action: Action):
+        self._action = action
+
+    async def run(self, drop: ffi.Drop):
+        try:
+            await self._action.run(drop)
+        except:
+            pass
```
LukasPukenis, line 116: This is more like `ExpectAnyErrorOrNoError`, it should fail if no error was produced I assume?

matszczygiel: Oh yeah, let me fix this

test/scenarios.py
```diff
@@ -6506,125 +6506,37 @@
...
-                                Error.IO,
-                            ),
-                        ]
+                            action.MakeHttpGetRequest(
+                                "http://172.20.0.15:49111/non-existing-path", 404
+                            )
+                        ],
+                        50,
+                    ),
+                    # Let's issure 10 more than the limit to ensure we triggered the protection
```
LukasPukenis, line 6519:
```suggestion
                    # Let's issure 10 more than the limit to ensure we triggered the protection
```
```suggestion
                    # Let's issue 10 more than the limit to ensure we triggered the protection
```

test/scenarios.py
```diff
@@ -6506,125 +6506,37 @@
...
-                            3,
-                            {
-                                event.File(
-                                    FILES["testfile-small"].id,
-                                    "testfile-small",
-                                    1048576,
-                                ),
-                            },
-                        )
+                    # Check if we get an apporpiate HTTP error code
```
LukasPukenis, line 6530:
```suggestion
                    # Check if we get an appropriate HTTP error code
```

LukasPukenis (approved):
As I understand we test by issuing manual HTTP requests towards our endpoint. This looks like a nice compromise but the real test with real transfers is more accurate as it may trigger DDoS protection differently. For the piece of mind I suggest leaving both new and old testcases, that's my 2ct

matszczygiel:
@LukasPukenis the old one was completely broken. I don't see a clear way to test this using another libdrop instance as a client.
But THB isn't that better that we test it with an HTTP client? In general malicious software may not be another libdrop instance

LukasPukenis:
@matszczygiel My point is that because we try to secure against our own app, we should test it as such as well to aim for the best results. Testing with HTTP client directly _should_ cover even more ground though so it's fine

PR #195, opened 2023-08-31
Fix all the started DB states contained 0 bytes transmitted

matszczygiel:
As in the title the `bytes_received` and `bytes_sent` field was not being set up in the DB. Even though it was not declared as `DEFAULT 0`

### DIFF
--- a/drop-transfer/src/storage_dispatch.rs
+++ b/drop-transfer/src/storage_dispatch.rs
@@ -20,14 +20,24 @@ impl<'a> StorageDispatch<'a> {
 
     pub async fn handle_event(&mut self, event: &crate::Event) {
         match event {
-            crate::Event::FileUploadStarted(transfer, file_id) => {
+            crate::Event::FileUploadStarted(transfer, file_id, bytes) => {
+                self.store_progres(transfer.id(), file_id, *bytes as _);
                 self.storage
-                    .insert_outgoing_path_started_state(transfer.id(), file_id.as_ref())
+                    .insert_outgoing_path_started_state(
+                        transfer.id(),
+                        file_id.as_ref(),
+                        *bytes as _,
+                    )
                     .await
             }
-            crate::Event::FileDownloadStarted(transfer, file_id, ..) => {
+            crate::Event::FileDownloadStarted(transfer, file_id, _, bytes) => {
+                self.store_progres(transfer.id(), file_id, *bytes as _);
                 self.storage
-                    .insert_incoming_path_started_state(transfer.id(), file_id.as_ref())
+                    .insert_incoming_path_started_state(
+                        transfer.id(),
+                        file_id.as_ref(),
+                        *bytes as _,
+                    )
                     .await
             }
             crate::Event::FileDownloadSuccess(transfer, download) => {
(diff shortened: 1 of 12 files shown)


drop-transfer/src/storage_dispatch.rs
```diff
@@ -20,14 +20,24 @@ impl<'a> StorageDispatch<'a> {
...
         match event {
-            crate::Event::FileUploadStarted(transfer, file_id) => {
+            crate::Event::FileUploadStarted(transfer, file_id, bytes) => {
+                self.store_progres(transfer.id(), file_id, *bytes as _);
                 self.storage
-                    .insert_outgoing_path_started_state(transfer.id(), file_id.as_ref())
+                    .insert_outgoing_path_started_state(
+                        transfer.id(),
+                        file_id.as_ref(),
+                        *bytes as _,
```
LukasPukenis, line 29: What does this do in Rust?

matszczygiel: Do you mean `as _` it's the same as `as i64` but the compiler can infer the type because of the argument type of the function

norddrop/src/device/types.rs
```diff
@@ -125,13 +125,13 @@ impl From<drop_transfer::Event> for Event {
         match e {
             drop_transfer::Event::RequestReceived(tx) => Event::RequestReceived(tx.as_ref().into()),
             drop_transfer::Event::RequestQueued(tx) => Event::RequestQueued(tx.as_ref().into()),
-            drop_transfer::Event::FileUploadStarted(tx, fid) => {
+            drop_transfer::Event::FileUploadStarted(tx, fid, _) => {
```
LukasPukenis, line 128: Adding new fields doesn't break compatibility, maybe worth adding them?

matszczygiel: Yeah, let me do it

drop-transfer/src/storage_dispatch.rs
LukasPukenis: oh okay, thanks

drop-transfer/src/ws/server/mod.rs
```diff
@@ -625,6 +625,8 @@ impl FileXferTask {
                 offset,
                 tmp_location,
             } => {
+                events.start(self.base_dir.to_string_lossy(), offset).await;
+
```
Lipt0nas, line 629: This is nice, we're emitting the event at a "more correct" time

Lipt0nas (dismissed):

matszczygiel:
@LukasPukenis Apparently the tests are correct everywhere, quite by accident I think

matszczygiel:
I'm going to wait for integration tests and then merge

matszczygiel:
@LukasPukenis I've implemented a check for the `transfer` filed in the `Started` event

LukasPukenis (approved):
I am a bit worried about that we do not always emit Start(Android testcase) event but the fix looks great otherwise

matszczygiel:
@LukasPukenis Can you explain when we do not emit the event?

LukasPukenis:
@matszczygiel scenarios.py contain Android testcase where the Start events are removed

PR #196, opened 2023-09-01
Periodic context updates

Lipt0nas:


drop-analytics/src/moose_impl.rs
```diff
@@ -110,17 +111,27 @@ impl MooseImpl {
...
+            loop {
+                populate_context(&task_logger);
+                interval.tick().await;
+            }
+        });
+
+        Ok(Self {
+            logger,
+            context_update_task,
+        })
```
matszczygiel, line 127: I think that this should be a normal `std` thread.
Since moose is static and our tokio runtime is not, I see a lot of things that can go wrong here.
Just use plain old OS thread

drop-transfer/src/manager.rs
```diff
@@ -44,13 +44,13 @@ pub enum FileTerminalState {
     Failed,
 }
 
-enum IncomingLocalFlieState {
+enum IncomingLocalFileState {
```
matszczygiel, line 47: Nice catch

drop-analytics/src/moose_impl.rs
Lipt0nas: I'm unaware of the potential problems that this could cause, could you explain a bit more?

matszczygiel: So moose object has a static lifetime hence it lives until the libdrop so is unloaded (potentially without running its destructor).
But tokio runtime lives up until `norddrop_destory()` is called.
So I expect disasters to happen after `norddrop_destory()` is called, like panics and potential crashes

matszczygiel: Moreover, you simply don't need tokio here. It's a simple problem to solve, the OS thread suffices.
Tokio is meant for heavy IO-bound things like servers etc.

Lipt0nas: Fair enough, changed it

drop-analytics/src/moose_impl.rs
```diff
@@ -110,17 +112,41 @@ impl MooseImpl {
...
+                    Ok(_) | Err(std::sync::mpsc::TryRecvError::Disconnected) => break,
+                    Err(std::sync::mpsc::TryRecvError::Empty) => {
+                        if last_update.elapsed() > Duration::from_secs(60 * 10) {
+                            populate_context(&task_logger);
+                            last_update = std::time::Instant::now();
+                        }
+                        std::thread::sleep(Duration::from_millis(100));
+                    }
+                }
+            }
```
matszczygiel, line 131: Oh, yeah. OS threads don't support cancelation.
But TBH should we care about the immediate cancelation?
IMHO it should look like this:
* we have an atomic bool `is_running` initialized to `true`
* we set it to `false` in the drop impl, and don't care about the joining
* in the thread we loop `while is_running`, populate the context, and then sleep for 10 min

In that way, we do not harm the battery usage with thread woken up every 100ms.
It also does not matter when we set up different moose because for sure the thread will still be alive but it won't do anything and will close itself eventually.
So one last thing is what happens when the app is closed, then the OS will simply kill the tread cleanly.

matszczygiel: I don't push hard for it but hope that I got you convinced

Lipt0nas: I'm not a fan of leaving the threads dangling around like this, but you do have a very convincing point from a battery perspective that I didn't take into account, so I went with your suggestion

matszczygiel (approved):

matszczygiel:
@Lipt0nas If you want to somehow terminate the thread immediately you can make use of something like this https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv_timeout with timeout set to 10mins upon which you update the context or terminate if the channel is closed

Lipt0nas:
Completely forgot about recv_timeout, even though it was used just a few lines above this code :facepalm:, the initial implementation was so close, just used the wrong damn recv :cry:

drop-analytics/src/moose_impl.rs
```diff
@@ -108,19 +113,33 @@ impl MooseImpl {
...
         populate_context(&logger);
-        let res = moose::flush_changes();
-        moose_debug!(logger, res, "flush_changes");
-        res.context("Failed to flush moose context")?;
 
-        Ok(Self { logger })
+        let (tx, rx) = sync_channel(1);
+        let task_logger = logger.clone();
+        let context_update_task = std::thread::spawn(move || loop {
+            match rx.recv_timeout(Duration::from_secs(60 * 10)) {
```
matszczygiel, line 121: Is this correct? I think that on timeout you would receive an error `Timeout` thus this loop will only turn once.
You should probably `populate_context()` on the error and break on `Ok()`.

BTW you don't need to send anything through the channel dropping it suffices, since you can infer it from the error https://doc.rust-lang.org/std/sync/mpsc/enum.TryRecvError.html

Lipt0nas: oh yeah, you're absolutely right on the logic, for the sending part, we'd still need to manually drop the channel to make sure that it happens before joining, so we can't save a code line and explicitly sending the signal looks a little more correct IMO

matszczygiel (approved):

PR #197, opened 2023-09-04
DB error triggered by android app on high latency network

matszczygiel:
The issue is that warp (WS server) does not guarantee message delivery on successful flushes - which is probably a bug on their side. This can cause a situation when the receiver thinks the close frame got delivered but in fact, it is not the case, leading to a mismatched transfer state.
Trying to receive from the socket after closing frame seems to catch the problem, though I don't think it catches it in 100% of cases.
Nonetheless, I managed to reproduce the issue with `scenario38` and managed to make it pass, so I don't know what else I could do to improve the implementation

### DIFF
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -138,7 +138,17 @@ impl TransferManager {
                 Ok(false)
             }
             Entry::Vacant(vacc) => {
-                self.storage.insert_transfer(&xfer.storage_info()).await;
+                if self
+                    .storage
+                    .insert_transfer(&xfer.storage_info())
+                    .await
+                    .is_none()
+                {
+                    warn!(self.logger, "Transfer was closed already");
+                    let _ = conn.send(ServerReq::Close);
+                    return Ok(false);
+                }
+
                 self.storage
                     .update_transfer_sync_states(
                         xfer.id(),
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -419,12 +419,9 @@ class ExpectCancel(Action):
 # Shape egress traffic. Slowing down and adding latency helps introducing
 # determinism in the testing environment
 class ConfigureNetwork(Action):
-    def __init__(
-        self, rate: str = "10mbit", latency: str = "3000ms", delay: str = "0ms"
-    ):
+    def __init__(self, rate: str = "10mbit", latency: str = "0ms"):
         self._rate = rate
         self._latency = latency
-        self._delay = delay
 
     async def run(self, drop: ffi.Drop):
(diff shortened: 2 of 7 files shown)


LukasPukenis:
Just leaving a comment about the additional solution of sending a CLOSE frame upon receiving the transfer with an already-existing ID

drop-transfer/src/manager.rs
```diff
@@ -138,7 +138,17 @@ impl TransferManager {
...
             }
             Entry::Vacant(vacc) => {
-                self.storage.insert_transfer(&xfer.storage_info()).await;
+                if self
+                    .storage
+                    .insert_transfer(&xfer.storage_info())
+                    .await
+                    .is_none()
+                {
+                    warn!(self.logger, "Transfer was closed already");
```
LukasPukenis, line 147: To be honest it was not closed, it was just there.. Closed would assume we have a Canceled state in DB

LukasPukenis (approved):

test/drop_test/action.py
```diff
@@ -436,12 +439,12 @@ def ex(cmd: str):
 
         device = "eth0"
         ex(
-            f"tc qdisc add dev {device} root tbf rate {self._rate} burst 64k latency {self._latency}"
+            f"tc qdisc add dev {device} root netem rate {self._rate} latency {self._latency} delay {self._delay}"
```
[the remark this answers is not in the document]
matszczygiel: Yeah, might be. But it does not work with tbf either.
Do you suggest unifying `latency` and `delay`?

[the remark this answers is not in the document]
matszczygiel: Let me experiment more with it

drop-transfer/src/manager.rs
LukasPukenis: Re-reading this code gives me more questions:
- We receive a transfer request
- We try to insert it
- If there was no such transfer, we insert it and check again. This looks like a bit unnecessary

- The message says the transfer was closed already but we have `cancel` in our semantics, so we should really check the state of the transfer if it contains a `cancel` state or else it just might be there not properly canceled as well

matszczygiel: Ok, I misused `closed` here. I meant `canceled`.
I don't agree with you. The source of truth is the hash map persisted in the form of sync tables. If there is no transfer in the hash map, it's not alive anymore (or does not exist), which is checked in this code.
Our logic does not make use of canceled state tables, these are for the clients of JSON API

LukasPukenis: Thanks for explaining. It makes sense now. I was confused for a short moment.

PR #198, opened 2023-09-04
Suppress unseen cancelled transfers syncing

matszczygiel:
This feature was requested by the Android team. They wanted to suppress showing the UI for transfers that are going to be canceled immediately

### DIFF
--- a/drop-transfer/src/ws/client/mod.rs
+++ b/drop-transfer/src/ws/client/mod.rs
@@ -5,6 +5,7 @@ mod v4;
 mod v5;
 
 use std::{
+    future::Future,
     io,
     net::{IpAddr, SocketAddr},
     ops::ControlFlow,
@@ -57,6 +58,13 @@ struct RunContext<'a> {
     xfer: &'a Arc<OutgoingTransfer>,
 }
 
+enum WsConnection {
+    Stopped,
+    Unrecoverable(crate::Error),
+    Recoverable,
+    Connected(WsStream, protocol::Version),
+}
+
 pub(crate) fn spawn(
     state: Arc<State>,
     xfer: Arc<OutgoingTransfer>,
@@ -83,18 +91,20 @@ pub(crate) fn spawn(
 
 async fn run(state: Arc<State>, xfer: Arc<OutgoingTransfer>, logger: Logger, alive: AliveGuard) {
     loop {
-        let cf = connect_to_peer(&state, &xfer, &logger, &alive).await;
-        if cf.is_break() {
-            if let Err(err) = state.transfer_manager.outgoing_remove(xfer.id()).await {
-                warn!(
-                    logger,
-                    "Failed to clear sync state for {}: {err}",
-                    xfer.id()
-                );
-            }
-
-            return;
+        if connect_to_peer(&state, &xfer, &logger, &alive)
+            .await
+            .is_break()
+        {
+            break;
         }
+    }
+
+    if let Err(err) = state.transfer_manager.outgoing_remove(xfer.id()).await {
+        warn!(
+            logger,
+            "Failed to clear sync state for {}: {err}",
+            xfer.id()
+        );
     }
 }
 
@@ -104,10 +114,11 @@ async fn connect_to_peer(
     logger: &Logger,
     alive: &AliveGuard,
 ) -> ControlFlow<()> {
-    let (socket, ver) = match establish_ws_conn(state, xfer.peer(), logger).await {
-        Ok(Some(res)) => res,
-        Ok(None) => return ControlFlow::Continue(()),
-        Err(err) => {
+    let (socket, ver) = match establish_ws_conn(state, xfer, logger).await {
+        WsConnection::Connected(sock, ver) => (sock, ver),
+        WsConnection::Recoverable => return ControlFlow::Continue(()),
+        WsConnection::Stopped => return ControlFlow::Break(()),
+        WsConnection::Unrecoverable(err) => {
             error!(logger, "Could not connect to peer {}: {}", xfer.id(), err);
 
             state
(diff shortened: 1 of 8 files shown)


LukasPukenis:
Hey, I've tried to review this PR but really couldn't grasp the solution behind it. Could you give a higher level overview of the solution, how and why it works?
I also have a bit of trouble understanding how this works in conjunction with https://github.com/NordSecurity/libdrop/pull/197

My thought process:
- If we know that we're canceling when emitting `TransferRequested` then we can omit that event entirely without introducing an extra field

I also suggest adding more info to the commits why Android is receiving events in such a manner and why it happens

drop-transfer/src/protocol/v5.rs
```diff
@@ -44,6 +44,8 @@ pub struct File {
 pub struct TransferRequest {
     pub files: Vec<File>,
     pub id: uuid::Uuid,
+    #[serde(default)]
```
LukasPukenis, line 47: What does `default` do in this case?

drop-transfer/src/protocol/v5.rs
```diff
@@ -196,6 +198,7 @@ impl From<&OutgoingTransfer> for TransferRequest {
                 })
                 .collect(),
             id: value.id(),
+            canceling: false,
```
LukasPukenis, line 201: This doesn't give me confidence... Should this field even be on this struct? It should probably not even reach this state if it's already "in cancel"

drop-transfer/src/storage_dispatch.rs
```diff
@@ -54,10 +54,7 @@ impl<'a> StorageDispatch<'a> {
                     .insert_outgoing_path_completed_state(transfer.id(), file_id.as_ref())
                     .await
             }
-            crate::Event::IncomingTransferCanceled(transfer, by_peer) => {
-                self.storage
-                    .insert_transfer_cancel_state(transfer.id(), *by_peer)
-                    .await;
+            crate::Event::IncomingTransferCanceled(transfer, _) => {
```
LukasPukenis, line 57: This `insert_transfer_cancel_state` is removed in this case?

LukasPukenis (commented):
Do we need the `canceling` field at all?

If the sender has a cancel in sync table and cancels state as well, we could have a state machine in there to decide if it's worth even sending the transfer over to peer. This way the logic would be contained in one place more or less separated from the main thing.

drop-transfer/src/protocol/v5.rs
matszczygiel: Ensures backward compatibility. We cannot expect all the peers to support the change so we need to assume that when this field is not present then it's set to false.
Of course, the proper way of doing this would be to introduce another protocol version, but I was too lazy for this

drop-transfer/src/protocol/v5.rs
matszczygiel: Um, why? It was reached by the Android team and reported as a problem

matszczygiel: So the thing is that you want to know as soon as possible that the canceling is happening

LukasPukenis: Yeah, but I am wondering if we should even send this xfer request if we know it's gonna be canceled. Elsewise we need to properly manage this `canceled` field which feels redundant if we already know it

drop-transfer/src/storage_dispatch.rs
matszczygiel: Yes, since it was based on the event. The event is suppressed in some cases so no state would be inserted.
To fix this I needed to insert the state independently of the event

drop-transfer/src/protocol/v5.rs
Lipt0nas: This is actually an interesting point - could we get away with just not communicating with the other side at all in this case? Since the whole reason for why Android wants this change is to not show anything on the UI, doubt they would need this info if they just want to instantly hide it

matszczygiel: @Lipt0nas do you suggest just skipping the transfer on the sender side altogether?
It's a valid option. However, in that case, you would have a mismatched history on both sides.
IDK, @LukasPukenis what do you think

Lipt0nas: @matszczygiel Yeah, that's the idea

I see only benefits in this approach, considering with this we would allow hiding transfers from the UI, as a receiver, I wouldn't want these "ghost" transfers to appear in my history, on a very big stretch we could even call it abuse, if i'm able to fill up the transfer history with meaningless entries :D

matszczygiel: That was the original solution proposed by the Android team. I opted against it, because of history.
But I guess you managed to change my mind.
Let me implement this the way you described, moving the PR to draft mode

matszczygiel:
@LukasPukenis @Lipt0nas I managed to implement the suppression of sending the unseen canceled transfers

test/scenarios.py
```diff
@@ -7874,4 +7872,87 @@
...
+        {
+            "ren": ActionList(
+                [
+                    action.SleepMs(400),
+                    action.Start("172.20.0.5"),
+                    action.NewTransfer("172.20.0.15", ["/tmp/testfile-small"]),
+                    action.Wait(
+                        event.Queued(
+                            0,
+                            {
```
Lipt0nas, line 7887: I guess the tests here are no longer valid, and by looks of it `scenario4-12` is doing this test, albeit it's completely broken :D

matszczygiel: Oh, yeah. Good point. Let me copy paste 39-2 into 4-12 and remove 31-1

drop-transfer/src/ws/client/mod.rs
```diff
@@ -57,6 +58,13 @@ struct RunContext<'a> {
     xfer: &'a Arc<OutgoingTransfer>,
 }
 
+enum WsConnection {
+    Stopped,
+    UnrecovError(crate::Error),
+    RecovFailure,
```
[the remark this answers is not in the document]
matszczygiel: Done

Lipt0nas (approved):
The solution and the diff became much cleaner, big nice

Maybe @LukasPukenis also wants to take a peek

LukasPukenis (approved):
Nice change and refactor :partying_face:

PR #199, opened 2023-09-07
Force recheck from the receiver side for active transfers

matszczygiel:
I've implemented two things:
* server endpoint for checking transfer state
* client loop issuing request for this endpoint on the receiver side of things

I've also added a test case, which sadly takes a lot of time to execute

### DIFF
--- a/drop-transfer/src/check.rs
+++ b/drop-transfer/src/check.rs
@@ -44,8 +44,19 @@ async fn run(state: Arc<State>, xfer: Arc<IncomingTransfer>, logger: Logger) {
     }
 
     info!(logger, "Transfer {} is gone. Clearing", xfer.id());
-    if let Err(err) = state.transfer_manager.incoming_remove(xfer.id()).await {
-        warn!(logger, "Failed to clear incoming transfer: {err:?}");
+
+    match state.transfer_manager.incoming_remove(xfer.id()).await {
+        Err(err) => {
+            warn!(logger, "Failed to clear incoming transfer: {err:?}");
+        }
+        Ok(false) => {
+            state
+                .event_tx
+                .send(crate::Event::IncomingTransferCanceled(xfer.clone(), true))
+                .await
+                .expect("Could not send a file cancelled event, channel closed");
+        }
+        _ => (),
     }
 }
 
--- a/drop-transfer/src/check.rs
+++ b/drop-transfer/src/check.rs
@@ -1,0 +1,118 @@
+use std::{net::SocketAddr, ops::ControlFlow, sync::Arc};
+
+use hyper::StatusCode;
+use slog::{debug, info, warn, Logger};
+use tokio_util::sync::CancellationToken;
+
+use crate::{auth, protocol, service::State, tasks::AliveGuard, IncomingTransfer, Transfer};
+
+pub(crate) fn spawn(
+    state: Arc<State>,
+    xfer: Arc<IncomingTransfer>,
+    logger: Logger,
+    guard: AliveGuard,
+    stop: CancellationToken,
+) {
+    let id = xfer.id();
+    let job = run(state, xfer, logger.clone());
+
+    tokio::spawn(async move {
+        let _guard = guard;
+
+        tokio::select! {
+            biased;
+
+            _ = stop.cancelled() => {
+                debug!(logger, "Check job stop for transfer: {id}");
+            },
+            _ = job => (),
+        }
+    });
+}
+
+async fn run(state: Arc<State>, xfer: Arc<IncomingTransfer>, logger: Logger) {
+    loop {
+        tokio::time::sleep(drop_config::ALIVE_CHECK_INTERVAL).await;
+
+        if !state.transfer_manager.is_incoming_alive(xfer.id()).await {
+            break;
+        }
+
+        if make_request(&state.auth, &xfer, &logger).await.is_break() {
+            break;
+        }
+    }
+
+    info!(logger, "Transfer {} is gone. Clearing", xfer.id());
+    if let Err(err) = state.transfer_manager.incoming_remove(xfer.id()).await {
+        warn!(logger, "Failed to clear incoming transfer: {err:?}");
+    }
+}
+
+async fn make_request(
+    auth: &auth::Context,
+    xfer: &IncomingTransfer,
+    logger: &Logger,
+) -> ControlFlow<()> {
+    let addr = SocketAddr::new(xfer.peer(), drop_config::PORT);
+    let client = hyper::Client::new();
+    let versions_to_try = [protocol::Version::V5];
+
+    for version in versions_to_try {
+        let url: hyper::Uri = format!("http://{addr}/drop/{version}/check/{}", xfer.id())
+            .parse()
+            .expect("URL should be valid");
+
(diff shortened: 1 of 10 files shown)


test/scenarios.py
```diff
@@ -7875,4 +7875,52 @@
...
+                                    "testfile-small",
+                                    1048576,
+                                ),
+                            },
+                        )
+                    ),
+                    action.SleepMs(200),
+                    action.Stop(),
+                    action.Start("172.20.0.5"),
+                    action.Sleep(61),  # 60s is a check request interval
```
LukasPukenis, line 7901: I had a hard time understanding how it stalls until I realized that we use :memory: for SQLite. Can we add explicit different storage paths so it would be apparent?

test/scenarios.py
```diff
@@ -7875,4 +7875,52 @@
...
+                            {
+                                event.File(
+                                    FILES["testfile-small"].id,
+                                    "testfile-small",
+                                    1048576,
+                                ),
+                            },
+                        )
+                    ),
+                    action.ExpectCancel([0], True),
```
LukasPukenis, line 7921: Additional test idea: have a regular download with the same database, stop and sleep for 59seconds, and then restart again to see if it doesn't accidentally remove the non-stale transfer

drop-transfer/src/check.rs
```diff
@@ -0,0 +1,129 @@
...
+        tokio::time::sleep(drop_config::ALIVE_CHECK_INTERVAL).await;
+
+        if !state.transfer_manager.is_incoming_alive(xfer.id()).await {
+            break;
+        }
+
+        if make_request(&state.auth, &xfer, &logger).await.is_break() {
+            break;
+        }
+    }
```
LukasPukenis, line 44: Could you explain this flow a bit? Why `make_request` is repeatedly called in the loop?

test/scenarios.py
```diff
@@ -7875,4 +7875,52 @@
...
+                ]
```
LukasPukenis, line 7922: Additional test: storage API with the stale transfer on both ends

LukasPukenis (commented):
A few things questions(might be wrong assumptions, lots of code):

- Why do we only check on the receiver side? It might very well be vice-versa scenario so both sides should be checking each other if they are aware of the transfer.

- Does storage API return a proper canceled state(maybe even failed) in case of the stale transfer? I can't observe clearly from the changes

drop-transfer/src/ws/server/mod.rs
```diff
@@ -79,176 +79,114 @@ struct StreamCtx<'a> {
...
-            .untuple_one()
-            .and(warp::ws())
-            .map(
-                move |version: protocol::Version, peer: SocketAddr, ws: warp::ws::Ws| {
+        };
+
+        let check_route = {
+            let nonces = nonce_store.clone();
+
+            base.and(warp::path!("check" / String))
```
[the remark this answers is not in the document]
matszczygiel: Yeah, you should thank warp developers. It's so rigid that adding a single endpoint took me like 4 hours while reading through error messages reminding me of C++ template errors.
To sum up. It was necessary for the code to even compile

test/scenarios.py
matszczygiel: I guess using on-disk storage for this test case is overkill but I can add a comment explaining

test/scenarios.py
matszczygiel: I'm not sure if I get it. Can you elaborate more on the test case flow?

drop-transfer/src/check.rs
matszczygiel: `make_request()` ping the sender to check if the transfer is alive. According to Slack discussion on the problem, the agreed-upon solution was to ping periodically the sender, hence the loop

test/scenarios.py
matszczygiel: Yeah, could be even embedded in this test

matszczygiel:
@LukasPukenis answering your questions:
* The WS client making a connection periodically is already a check
* Yeah, might be worth adding test cases for this. However, this should work fine since it emits a event, and storage works by intercepting the events

drop-transfer/src/ws/server/mod.rs
```diff
@@ -79,176 +79,114 @@ struct StreamCtx<'a> {
     events: &'a FileEventTx<IncomingTransfer>,
 }
 
+#[derive(Debug)]
+struct MissingAuth(SocketAddr);
+impl warp::reject::Reject for MissingAuth {}
+
+#[derive(Debug)]
+struct Unauthrorized;
```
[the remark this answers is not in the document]
matszczygiel: done

test/scenarios.py

test/scenarios.py
matszczygiel: done

drop-transfer/src/ws/server/mod.rs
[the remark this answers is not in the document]
LukasPukenis: Ok, understandable, if it's justified then it's better :)

drop-transfer/src/check.rs
LukasPukenis: I understand that the solution is to periodically ping the peer with `GET /check` to see if the transfer is still alive but am wondering about the code readability. From my perspective, I would expect:

```
make_request().await
loop {
  if !state.transfer_manager.check_if_alive(xfer.id()).await {
     ...
}
```

test/scenarios.py
LukasPukenis: Sure. I don't think it is necessary thinking about it twice, my point was - try to stop and restart without triggering the deadline but there should be plenty of tests already doing start/stop/start with the same database

LukasPukenis:
@matszczygiel 
- What is the status of stale transfer removal? Is it considered canceled by a peer?
- "The WS client making a connection periodically is already a check". I suppose by that you mean that the receiver pings the sender with `GET /check` but the sender doesn't ping the receiver as it periodically will try to connect with a transfer and if it's not present on the receiver side it will provide an error?

test/scenarios.py
matszczygiel: Ok, if you say so I will skip this one

matszczygiel:
@LukasPukenis 
* yes, the status is cancelled by the peer
* correct

LukasPukenis:
@matszczygiel After a call this is the conclusion:
- I was confused about `make_transfer()` as it's present in check.rs and client.rs files due to clicking on the function definition in GitHub
- Sender periodically pings with transfers that are still alive and will fail if the peer responds with an error
- Receiver issues GET /check to the sender periodically and will produce a cancel when appropriate

All of the above is good.

We agreed on an additional test case where the transfer is initiated and we sleep for a minute to trigger the periodic check and then we should expect nothing to happen to the transfer on both ends.

@Lipt0nas might want to take a look as well

Lipt0nas:
In this case if we were to ping a non-v5 peer, it would error out?

matszczygiel:
@Lipt0nas not really. It would see the error status code and would not take any action. It would just go back to the beginning of the loop.

LukasPukenis (approved):

PR #201, opened 2023-09-11
Fix `FileDone` and `TransferCancelation` message reordering

matszczygiel:
Is blocked by https://github.com/NordSecurity/libdrop/pull/199

The commit message describes it fully. 
I couldn't reproduce the issue. However, I do see a logic flaw which can lead to it.
There are plenty of test cases that should exercise this (file transferred, then transfer cancelation) and none of them fails spectacularly, suggesting that the issue is super rare to observe

### DIFF
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -867,6 +867,9 @@ impl IncomingState {
 
                     Some(ServerReq::Fail {
                         file: file_id.clone(),
+                        msg: String::from(
+                            "File failed. The failed state was retrieved from the database.",
+                        ),
                     })
                 }
                 _ => None,
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 This fixes the case when transfer cancelled on the sender side caused `TransferCancelled` being fired immedieatly after `TransferRequest`.\
 With this change the transfer would not be seen on the receiver side at all.
 * Implement a periodic transfer state check in case of stalled transfer for the receiver
+* Fix occasional events reordering, for example file done swapped transfer cancelation, on the sender side 
 
 ---
 <br>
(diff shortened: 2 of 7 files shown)


drop-transfer/src/manager.rs
```diff
@@ -867,6 +867,7 @@ impl IncomingState {
 
                     Some(ServerReq::Fail {
                         file: file_id.clone(),
+                        msg: String::from("File failed elsewhere"),
```
LukasPukenis, line 870: :D could we provide a better worded hint?

LukasPukenis: we can add in additional context like "it might be due to x, y or z

LukasPukenis (approved):
I am assuming this is just reordering stuff

matszczygiel:
@LukasPukenis more or less, but it used different mechanisms for transmitting `Start` and `Done` messages so that they are not reordered with for example transfer cancelation

drop-transfer/src/manager.rs
matszczygiel: Um, the reason for this failure may literally be anything. The fun part is that this message is copied and was introduced before this PR :)
I will try to reword it a bit

matszczygiel: done

drop-transfer/src/manager.rs
```diff
@@ -867,6 +867,9 @@ impl IncomingState {
 
                     Some(ServerReq::Fail {
                         file: file_id.clone(),
+                        msg: String::from(
+                            "File failed. The failed state was retreived from the database.",
```
LukasPukenis, line 871: `retrieved`

LukasPukenis (dismissed):

PR #203, opened 2023-09-12
File 404/parallelize test runner

LukasPukenis:
Basically, there's one big commit I found trouble splitting into self-contained parts, it's committed message describes the inner stuff.

**TLDR**:
- tests are executed in parallel with `--scenarios-at-once` flag. Default value is bigger than existing scenario count.
- test files moved to docker image as `dd` was super slow in parallel that it actually froze my machine several times
- ipv6 test and special file symbol tests are failing, some tests are flaky but it's usual in our case.
- All test cases run in a pretty short amount of time(CI took just above 3 minutes)

Given that there are no glaring issues, I will fix the remaining issues.


matszczygiel:
I have a suspicion as to why the IPv6 test does not work. There are two networks in the container, one IPv4 and one IPv6.
For the IPv6 tests the socket is connecting to the IPv6 peer but I suspect they themselves bind to the IPv4 adapter, thus they cannot reach the peer.
Previously there was only one network in the container thus they were binding correctly

test/Makefile
```diff
@@ -15,11 +18,14 @@ build:
 
 # Build the project with coverage instrumentation
 build-coverage:
-	CARGO_INCREMENTAL=0 RUSTFLAGS="$(RUSTFLAGS_COVERAGE)" cargo build $(FEATURES)
+	CARGO_INCREMENTAL=0 RUSTFLAGS="$(RUSTFLAGS_COVERAGE)" cargo +nightly build $(FEATURES)
 
 # Test the project
 test: build
```
matszczygiel, line 24: Can you include the `build-image` stage or something that builds the docker image automatically?

test/Makefile
```diff
@@ -15,11 +18,17 @@ build:
 
 # Build the project with coverage instrumentation
 build-coverage:
-	CARGO_INCREMENTAL=0 RUSTFLAGS="$(RUSTFLAGS_COVERAGE)" cargo build $(FEATURES)
+	CARGO_INCREMENTAL=0 RUSTFLAGS="$(RUSTFLAGS_COVERAGE)" cargo +nightly build $(FEATURES)
+
+build-image:
+	docker build -t libdroptestimage test
```
matszczygiel, line 24:
```suggestion
	docker build -t libdroptestimage .
```

test/Makefile
```diff
@@ -15,11 +18,17 @@ build:
...
+build-image:
+	docker build -t libdroptestimage .
 
 # Test the project
-test: build
-	LIB_PATH=$(LIB_DOCKER) ./runner.py
+test: build build-image
+	docker kill $(docker ps -q) || true
+	docker network prune -f
+	docker container prune -f
```
matszczygiel, line 30: Can you include this one in the coverage as well? Possibly extracting it into a separate rule

LukasPukenis: sure!

LukasPukenis: added

matszczygiel (approved):
Good job :+1:

PR #207, opened 2023-09-13
Fix runner for ipv6

LukasPukenis:
Found that Peer is unused and simplified code by using OS calls for resolving. scenario28 now passes properly


matszczygiel (approved):
:+1:

PR #208, opened 2023-09-13
Use 0600 permissions for the database file on unix systems

matszczygiel:


drop-storage/src/lib.rs
```diff
@@ -32,9 +32,50 @@ pub struct Storage {
 
 const MIGRATIONS_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/migrations");
 
+#[cfg(unix)]
+fn prepare_sqlite_file(path: &str) -> io::Result<OpenFlags> {
```
LukasPukenis, line 36: The name is misleading, it doesn't prepare the file in any way, but builds the flags instead. `get_sqlite_flags()` or `build_sqlite_flags()` can be an alternative

drop-storage/src/lib.rs
```diff
@@ -32,9 +32,50 @@ pub struct Storage {
...
+                .write(true)
+                .create_new(true)
+                .mode(MODE)
+                .open(path)?;
+        }
+        Err(err) => return Err(err),
+    }
+
+    Ok(flags)
+}
```
LukasPukenis, line 68: We should include a test case for permissions if we alter the existing one as well, just in case. Initially, I thought about just affecting newly created database permissions and leaving existing ones as it is as it's additional problem vector. What is your take on this?

drop-storage/src/lib.rs
matszczygiel: Um, it creates a file and changes it's permissions. I think `get_sqlite_flags()` would be misleading since it doesn't mention messing around with the file system

drop-storage/src/lib.rs
```diff
@@ -32,9 +32,50 @@ pub struct Storage {
...
+
+#[cfg(not(unix))]
+fn prepare_sqlite_file(path: &str) -> io::Result<OpenFlags> {
+    Ok(OpenFlags::default())
+}
```
LukasPukenis, line 73: Windows should have teir own 0600 implementation. I think we could even pass it directly and it should just work

drop-storage/src/lib.rs
LukasPukenis: My bad, I started to comment without reading the body of the function

drop-storage/src/lib.rs
matszczygiel: Sadly it doesn't. At least in the Rust standard library

drop-storage/src/lib.rs
matszczygiel: Yeah. I can add those test cases

drop-storage/src/lib.rs
matszczygiel: Let me research this :+1:

drop-storage/src/lib.rs
matszczygiel: done

drop-storage/src/lib.rs
matszczygiel: So, as mentioned in today's sync, we leaving the Windows DB permissions for now

LukasPukenis (approved):
Beautiful testcases

PR #209, opened 2023-09-14
Lukasp/flaky testcase 31 1

LukasPukenis:


test/drop_test/peer_resolver.py
```diff
@@ -4,31 +4,9 @@
...
-        return self._name
-
-    def get_scenario(self) -> str:
-        return self._scenario
-
-
 class PEERResolver:
     def __init__(self):
         self._peer_mappings = {}
-        self._cache = {}
```
Lipt0nas, line 31: Could you elaborate on why was the caching removed?

LukasPukenis: Sure! So I noticed that caching is unnecessary as we can directly probe OS for reverse DNS lookup. Caching works fine as well however it makes it strict that a peer must be first queried with its name. I just refactored code and it seemed to work while being simpler to manage

PR #212, opened 2023-09-15
Fix: sender emits file finished each time it connects to the receiver

matszczygiel:
I noticed that the `TransferManager::outgoing_terminal_recv()` already reports that the file is already terminated, just the code inside the clients was simply ignoring it. Hence the event was emitted even though the file was already terminated.

I managed to do a little™ refactor :smile: as well.

I think it best to review this PR commit by commit

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -8353,4 +8353,73 @@ scenarios = [
             ),
         },
     ),
+    Scenario(
+        "scenario43",
+        "Sucesfully transfer the file. Then restart the receiver. Expect no events on receiver reconnection",
+        {
+            "DROP_PEER_REN": ActionList(
+                [
+                    action.Start("DROP_PEER_REN"),
+                    action.NewTransfer("DROP_PEER_STIMPY", ["/tmp/testfile-small"]),
+                    action.Wait(
+                        event.Queued(
+                            0,
+                            {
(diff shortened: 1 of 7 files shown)


LukasPukenis:
What do you think about introducing a separate testcase just for this? 31-8 is for nested directories. It's easier to understand what's wrong when a very specific test case fails. Or we can just modify existing one adding yet another stop()+start()+expect(noEvent)

matszczygiel:
@LukasPukenis oh yeah, let me add the test case

matszczygiel:
@LukasPukenis done

test/scenarios.py
```diff
@@ -8353,4 +8353,73 @@
...
+                        )
+                    ),
+                    action.Wait(event.Start(0, FILES["testfile-small"].id)),
+                    action.Wait(
+                        event.FinishFileUploaded(
+                            0,
+                            FILES["testfile-small"].id,
+                        )
+                    ),
+                    action.NoEvent(2),
```
LukasPukenis, line 8383: nitpick: this test might use a bit bigger NoEvent argument, especially when parallel executing it might be that receiver will do it's action slower and the sender will already succeed at waiting 2 seconds with no events

LukasPukenis (approved):

PR #213, opened 2023-09-18
Recompile library only when rs file is changed

LukasPukenis:


matszczygiel:
To be entirely correct, you need to check also `.toml`, `.lock`, and `.sql` files

matszczygiel (approved):

LukasPukenis:
@matszczygiel you're right, thanks! Updated it.

PR #215, opened 2023-09-18
Lukasp/trim testrunner

LukasPukenis:


test/runner.py
```diff
@@ -27,15 +27,6 @@ def prepare_docker() -> docker.DockerClient:
     client = docker.DockerClient(base_url="unix://var/run/docker.sock")
 
     # Network creation
-    ipv4_net = client.networks.create(
```
matszczygiel, line 30: Why it's unnecessary?

LukasPukenis: there's `net6` just few lines below, so this wasn't necessary

matszczygiel: But it doesn't contain an IPv4 range. How is it possible that it still works?

LukasPukenis: Basically because net6 contains both :)

matszczygiel (approved):

PR #216, opened 2023-09-18
tests: Fix scenario19-2

matszczygiel:


LukasPukenis (approved):
could you elaborate on how this fixes the test case? I looked at it but couldn't think why it would fail with the existing approach, the file appeared as expected in docker image..

matszczygiel:
The script was producing the file with literally `\x0A` in the name. Meaning it was not interpreted as a single line feed character. I've changed the script to produce a filename with the actual line feed inside

PR #217, opened 2023-09-19
Implement setting up the TCP SO_LINGER flag on the WS client

matszczygiel:


LukasPukenis:
The commit message could really help from the description why this was done

matszczygiel:
@LukasPukenis I've improved the commit message

LukasPukenis:
The commit message is better but I think it's still not trivial to understand what happens without this change. For example, if I would remove this code, all the tests would still pass, how could I ensure that it actually fixes something?

Can we somehow enforce a bug without this fix and make this fix solve the testcase?

Reading about SO_LINGER doesn't make it easier as it mentions that it will try to guarantee that data is being sent for the specified timeout instead of immediately closing the socket(which doesn't really happen anyway as Kernel deals with this detail).

matszczygiel:
@LukasPukenis The flag is for the kernel, so it should enforce some behavior.
TBH I don't quite understand how it works and the solution was suggested to me rather than it was my idea.
If you're not sure about this, we can abandon the PR altogether since in reality it doesn't solve any specific issue but is meant as a general improvement, whatever that means

LukasPukenis:
I am a bit divided, it's an honest effort so I wouldn't like for it to go to waste but at this state it's a liability(if some edge case would occur because of this it would be hard to track down)

matszczygiel:
@LukasPukenis Yeah, I totally understand it. Let's close the PR then. In any case we can always reopen it again

PR #218, opened 2023-09-19
[Draft] Preliminary moose API integration

Lipt0nas:
Just in case I decide to Thanos snap my changes away


drop-analytics/src/mock_impl.rs
```diff
@@ -1,37 +1,13 @@
...
-        _name: String,
-    ) {
-    }
+    fn event_init(&self, data: crate::InitEventData) {}
+    fn event_transfer_intent(&self, data: crate::TransferIntentEventData) {}
+    fn event_transfer_start(&self, data: crate::TransferStartEventData) {}
+    fn event_transfer_end(&self, data: crate::TransferEndEventData) {}
+    fn event_transfer_file(&self, data: crate::TransferFileEventData) {}
+    fn developer_exception(&self, data: crate::DeveloperExceptionEventData) {}
+    fn developer_exception_with_value(&self, data: crate::DeveloperExceptionWithValueEventData) {}
```
matszczygiel, line 12: Minor:
```suggestion
    fn event_init(&self, data: crate::InitEventData) {}
    fn event_transfer_intent(&self, _: crate::TransferIntentEventData) {}
    fn event_transfer_start(&self, _: crate::TransferStartEventData) {}
    fn event_transfer_end(&self, _: crate::TransferEndEventData) {}
    fn event_transfer_file(&self, _: crate::TransferFileEventData) {}
    fn developer_exception(&self, _: crate::DeveloperExceptionEventData) {}
    fn developer_exception_with_value(&self, _: crate::DeveloperExceptionWithValueEventData) {}
```
and remove the `allow(unused_`

drop-transfer/src/error.rs
```diff
@@ -111,14 +112,14 @@ impl From<&Error> for i32 {
...
 }
 
 impl<T> ResultExt for super::Result<T> {
-    fn to_status(&self) -> Result<(), i32> {
+    fn to_status(&self) -> i32 {
         match self {
-            Ok(_) => Ok(()),
-            Err(err) => Err(u32::from(err) as _),
+            Ok(_) => MOOSE_STATUS_SUCCESS,
+            Err(err) => u32::from(err) as _,
```
matszczygiel, line 122: Since it's now explicitly using moose status codes, maybe it's worth renaming it to something like `to_moose_status()`? So that it's not used by accident in some other context

drop-transfer/src/service.rs
```diff
@@ -54,6 +55,7 @@ impl Service {
         config: Arc<DropConfig>,
         moose: Arc<dyn Moose>,
         auth: Arc<auth::Context>,
+        init_time: Option<Instant>,
```
matszczygiel, line 58: Is `init_time` needed? I think we can always set `init_time` to `Instant::now()` inside this function. I know there are a couple of steps before this function is called but these things are not doing any network IO or something so the delay is pretty negligible

drop-transfer/src/ws/client/mod.rs
```diff
@@ -64,7 +65,7 @@ enum WsConnection {
     Stopped,
     Unrecoverable(crate::Error),
     Recoverable,
-    Connected(WsStream, protocol::Version),
+    Connected(WsStream, protocol::Version, i32),
```
matszczygiel, line 68: Let's use named fields for this one since it's unclear what `i32` represents

drop-transfer/src/ws/client/mod.rs
```diff
@@ -275,19 +284,20 @@ async fn tcp_connect<Fut: Future<Output = bool>>(
     remote_ip: IpAddr,
     break_check: impl Fn() -> Fut,
     logger: &Logger,
-) -> Option<TcpStream> {
+) -> Option<(TcpStream, i32)> {
```
matszczygiel, line 287: Or maybe we can do `type RetryCount = i32` for it or `struct RetryCount(i32)`

drop-transfer/src/error.rs
Lipt0nas: good point, will do

drop-transfer/src/ws/client/mod.rs
```diff
@@ -275,19 +284,20 @@ async fn tcp_connect<Fut: Future<Output = bool>>(
...
     let remote = SocketAddr::new(remote_ip, drop_config::PORT);
     let local = SocketAddr::new(local_ip, 0);
 
     let mut sleep_time = Duration::from_millis(200);
+    let mut retry_count = 0;
 
     loop {
```
matszczygiel, line 294:
```suggestion
    for retry_count in 0.. {
```

drop-transfer/src/service.rs
Lipt0nas: yeah, in this case it is needed, during the talk where we decided to introduce this metric, the two main points were because of the database opening and moose, which are the two main things that happen before going into service

drop-transfer/src/ws/events.rs
```diff
@@ -126,21 +162,24 @@ impl<T: Transfer> FileEventTx<T> {
...
+                transferred: utils::to_kb(lock.transferred),
+                result: status as _,
+            });
         }
     }
 }
 
 impl FileEventTx<IncomingTransfer> {
     pub async fn progress(&self, transfered: u64) {
+        self.inner.write().await.transferred = transfered;
```
matszczygiel, line 182: I don't this is correct. The `emit()` function locks on its own, so I suggest adjusting this function so that the locking takes place only once here.

norddrop/src/device/mod.rs
```diff
@@ -665,12 +669,12 @@ fn open_database(
                     Ok(storage) => Ok(storage),
                     Err(err) => {
                         let error = ffi::types::NORDDROP_RES_DB_ERROR;
-                        moose.developer_exception(
-                            error as i32,
-                            err.to_string(),
-                            "Failed to open DB after cleanup".to_string(),
-                            "DB Error".to_string(),
-                        );
+                        moose.developer_exception(DeveloperExceptionEventData {
```
matszczygiel, line 672: Maybe the `DeveloperExceptionEventData` should contain `&str`? Since most of the things here are static strings

test/prepare_files.sh
```diff
@@ -35,6 +35,12 @@ dd bs=1024K count=1 if=/dev/urandom of="/tmp/duplicate/testfile-small"
 dd bs=1024K count=1 if=/dev/urandom of="/tmp/duplicate/testfile.small.with.complicated.extension"
 dd bs=10240K count=2 if=/dev/urandom of="/tmp/duplicate/testfile-big"
 
+# Create tiny jpeg and gif files to test analytics
+# "/tmp/tiny-jpeg.jpg"
+echo -ne "\xFF\xD8\xFF\xE0\x00\x10\x4A\x46\x49\x46\x00\x01\x01\x01\x00\x48\x00\x48\x00\x00\xFF\xDB\x00\x43\x00\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xC2\x00\x0B\x08\x00\x01\x00\x01\x01\x01\x11\x00\xFF\xC4\x00\x14\x10\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xFF\xDA\x00\x08\x01\x01\x00\x01\x3F\x10" > "/tmp/tiny-jpeg.jpg"
+# "/tmp/tiny-gif.gif"
+echo -ne "\x47\x49\x46\x38\x39\x61\x01\x00\x01\x00\x00\xff\x00\x2c\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x00\x3b" > "/tmp/tiny-gif.gif"
```
matszczygiel, line 42: Wow, nice. I would not think of trying to test with real files ;)

drop-transfer/src/service.rs
matszczygiel: Ok :+1:

norddrop/src/device/mod.rs
Lipt0nas: the moose functions still expect a `String` in the end, while this would technically be an optimization when using the other moose implementations, i chose to keep it consistent with the other structs

matszczygiel: Ok. I get your point since real moose is used in the prod, we should optimize for it

drop-transfer/src/ws/events.rs
```diff
@@ -49,17 +53,25 @@ impl FileEventTxFactory {
...
 impl<T: Transfer> FileEventTx<T> {
-    fn file_info(&self) -> Option<FileInfo> {
-        Some(self.xfer.files()[&self.file_id].info())
+    fn file_info(&self) -> FileInfo {
+        self.xfer.files()[&self.file_id].info()
     }
 
     async fn emit(&self, event: Event) {
-        let lock = self.inner.read().await;
+        let mut lock = self.inner.write().await;
```
matszczygiel, line 61: Now it's always writing. In that case, I think it should be changed to `tokio::sync::Mutex`

matszczygiel:
I'm not approving it since it is still a draft, but in the current state LGTM :+1:

drop-transfer/src/service.rs
```diff
@@ -199,12 +204,23 @@ impl Service {
 
     pub async fn send_request(&mut self, xfer: crate::OutgoingTransfer) {
         let xfer = Arc::new(xfer);
+
+        self.state.moose.event_transfer_intent(xfer.info());
```
matszczygiel, line 208: `intent`? Is it a typo?

Lipt0nas: intent is the correct name of the event, the word roughly means the wish or the intention of something, and this events describes just that - the intention to send an event, before any other action like connection is made

test/scenarios.py
```diff
@@ -9874,4 +9758,802 @@
             ),
         },
     ),
+    Scenario(
+        "scenario-analytics-1",
```
matszczygiel, line 9762: I think it's better to name them the same, but at the same using tags

Lipt0nas: good point, forgot about the tags, added them

drop-transfer/src/service.rs
matszczygiel: I've checked the code and it seems that the `init_time` can be `Instant` (without option). The `None` value is used in udrop only but since it's a demo, there's no need to optimize the API for it and we can simply use `Instant::now()` in the udrop

matszczygiel (changes requested):
I'd like to see the change from the `init_time` comment

drop-transfer/src/service.rs
Lipt0nas: Good point, done

matszczygiel (approved):

test/prepare_files.sh
```diff
@@ -42,6 +42,12 @@ dd bs=10240K count=2 if=/dev/urandom of="/tmp/duplicate/testfile-big"
 dd bs=1024K count=1 if=/dev/urandom of="/tmp/name/file-01"
 dd bs=1024K count=1 if=/dev/urandom of="/tmp/different/name/file-02"
 
+# Create tiny jpeg and gif files to test analytics
+# "/tmp/tiny-jpeg.jpg"
+echo -ne "\xFF\xD8\xFF\xE0\x00\x10\x4A\x46\x49\x46\x00\x01\x01\x01\x00\x48\x00\x48\x00\x00\xFF\xDB\x00\x43\x00\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xC2\x00\x0B\x08\x00\x01\x00\x01\x01\x01\x11\x00\xFF\xC4\x00\x14\x10\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xFF\xDA\x00\x08\x01\x01\x00\x01\x3F\x10" > "/tmp/tiny-jpeg.jpg"
+# "/tmp/tiny-gif.gif"
+echo -ne "\x47\x49\x46\x38\x39\x61\x01\x00\x01\x00\x00\xff\x00\x2c\x00\x00\x00\x00\x01\x00\x01\x00\x00\x02\x00\x3b" > "/tmp/tiny-gif.gif"
```
LukasPukenis, line 49: that's nice

matszczygiel:
BTW @Lipt0nas shouldn't this land in `dev`? I'm not sure

changelog.md
```diff
@@ -1,3 +1,8 @@
+### v6.0.0
+### **UNRELEASED**
+---
+* Update moose tracker to v2.0.0
+
```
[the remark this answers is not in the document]
LukasPukenis: Assuming we can merge painlessly into `main`. This woud be nicer as current `dev` has my new API which is incomplete

LukasPukenis (approved):
Great effort

PR #220, opened 2023-09-19
Fix: the docker network can reuse ip address between the containers failing some of the scenarios

matszczygiel:
Upon inspecting the containers I found that the MAC address is being reused across the containers (they are reassigned even between stopping and starting the container).
I decided to assign MAC addresses incrementally since that was causing the IP reuse in the first place (all the sane DHCP implementations won't share IPs between different MAC addresses, I guess).

### DIFF
--- a/test/runner.py
+++ b/test/runner.py
@@ -36,11 +36,16 @@ def prepare_docker() -> docker.DockerClient:
                 docker.types.IPAMPool(
                     subnet="fd3e:0e6d:45fe:b0c2::/64", gateway="fd3e:0e6d:45fe:b0c2::1"
                 )
-            ]
+            ],
+            options={"strictAffinity": "true"},
         ),
     )
 
     return client
+
+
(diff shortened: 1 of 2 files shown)


test/runner.py
```diff
@@ -36,13 +36,18 @@ def prepare_docker() -> docker.DockerClient:
...
         ),
     )
 
     return client
 
 
+def int_to_mac(macint: int):
+    return ":".join(re.findall("..", "%012x" % macint))
+
+
```
LukasPukenis, line 50: that's a neat trick :D

LukasPukenis (approved):
good find!

test/runner.py
matszczygiel: copied from the internet :D

matszczygiel:
@LukasPukenis I will convert this back to draft since I think I just witnessed IP reuse again. Need to investigate

PR #221, opened 2023-09-19
norddrop: Disable rlib builds

ernestask:
Building rlibs breaks on Windows due to winresource setting the +whole-archive link modifier. This commit disables rlib builds for lack of a better solution.

### DIFF
--- a/norddrop/Cargo.toml
+++ b/norddrop/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Lukas Pukenis"]
 edition = "2021"
 
 [lib]
-crate-type = ["staticlib", "cdylib", "lib"]
+crate-type = ["staticlib", "cdylib"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


matszczygiel:
@ernestask I think the `rlib` is used by the Linux app team. Does this change suit them?

ernestask:
> @ernestask I think the `rlib` is used by the Linux app team. Does this change suit them?

Where is it used? They produce staticlibs from libtelio and libdrop.

LukasPukenis:
Linux app team compile Libdrop from Rust source

LukasPukenis (approved):

Lipt0nas:
It would in fact break the Linux build, all of the imports would not be visible with these changes, but @ernestask mentioned that the upcoming rust version bump might introduce more problems

Let's keep this MR here until the bump actually happens and go from there

PR #223, opened 2023-09-20
Fix: sending identical named folders from different paths end up as a single folder with contents merged

matszczygiel:
This was reported as a bug. So here is the fix.
I've implemented the dedicated file-gathering methods so that it's easier to inspect and reuse those.
The logical change is recording file names in the hash set and checking if the name already exists. If it does then (1), (2),... suffix is added to the subpath's first component.

I'm not sure what to do with the file (not directories) names, it makes 23-1 failing. What do you think?

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -3034,6 +3034,126 @@ scenarios = [
         },
     ),
     Scenario(
+        "scenario15-3",
+        "Send two different directories with the same name, their content should not being merged",
+        {
+            "DROP_PEER_REN": ActionList(
+                [
+                    action.WaitForAnotherPeer(),
+                    action.Start("DROP_PEER_REN"),
+                    action.NewTransfer(
+                        "DROP_PEER_STIMPY", ["/tmp/name", "/tmp/different/name"]
+                    ),
+                    action.Wait(
+                        event.Queued(
+                            0,
+                            {
+                                event.File(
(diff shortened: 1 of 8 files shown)


test/scenarios.py
```diff
@@ -3033,6 +3033,126 @@
             ),
         },
     ),
+    Scenario(
+        "scenario15-3",
+        "Send two different directories with the same name, their content not being merged",
```
Lipt0nas, line 3038:
```suggestion
        "Send two different directories with the same name, their content should not being merged",
```

Lipt0nas (approved):

test/scenarios.py

matszczygiel:
@Lipt0nas @LukasPukenis it's still not clear what to do with same-named files (not dirs)

Lipt0nas:
oh i didn't see that it changed the logic for for single file transfers as well, i guess the idea would be to apply this new logic only to files in a directory, while single files should operate as they did (not checking or inserting into the hash set), though this seems like it might not be a clean looking solution

matszczygiel:
Yeah, probably that's the safest way to go. Let me change this

matszczygiel:
@Lipt0nas I think it's done. Files are not renamed anymore - same as before

Lipt0nas (approved):
Nice, i like how the `file/mod.rs` became cleaner now as well

PR #227, opened 2023-09-25
Disallow empty transfers

Lipt0nas:


test/scenarios.py
```diff
@@ -3353,6 +3353,28 @@
...
+    Scenario(
+        "scenario17-1",
+        "Try to send an empty directory, expect error",
+        {
+            "DROP_PEER_REN": ActionList(
+                [
+                    action.Start("DROP_PEER_REN"),
+                    action.ConfigureNetwork(),
+                    action.WaitForAnotherPeer(),
+                    action.NewTransferFails("DROP_PEER_STIMPY", "/tmp/empty-dir"),
```
LukasPukenis, line 3365: Can we add a nested deep directory as well:
```
/tmp/empty-dir
/tmp/empty-dir/dir/
/tmp/empty-dir/dir/dir
```

and upon sending `/tmp/empty-dir` expect an error?

LukasPukenis (dismissed):
nice!

test/scenarios.py
LukasPukenis: Just an additional check

Lipt0nas: Done

LukasPukenis (approved):

PR #230, opened 2023-09-27
FILE-501 peer online signal

LukasPukenis:
Introduce `norddrop_set_peer_state(peer, state)` API
---
- removed logic for infinite loop with exponential, capped sleep
- transfer is initiated when:
  * libdrop starts
  * A new transfer is made
  * norddrop_set_peers_state() with `online` flag is passed

- Each testcase became very sensitive to timing as without automated retry if the transfer is initiated from A->B, B is online but libdrop has not yet started it will then return `CONN REFUSED` immediately and without retry logic it will fail. For this, the solution was to add `WaitForAnotherPeer()` in each testcase. An alternative could be to add `norddrop_set_peers_state()` but I didn't see a reason for that.
- `offline` state is actually ignored but I think it's nice to ask for it from the apps as they have that state ready, however not a hard opinion on this, we can rework it into `set_peers_online(json_array_of_peers)`. Opinions are welcome :)
- all offline testcases(those where one of the peers stop libdrop) now contains explicit `norddrop_set_peers_state()` call to inform that peer is online
- tags were introduced in scenarios for easy running via `TAGS=offline,reject make -C test`
- Simplified DDoS protection logic as previously if it was tripped it would remove the transfer. As we moved the connection to a manual API call, it's easy to trip on it accidentally and remove the transfer thus I reworked DDoS protection to be recoverable error.
- Added rapidly disconnecting peer scenarios(35-*, 44-2, 44-3).

### DIFF
--- a/drop-analytics/Cargo.toml
+++ b/drop-analytics/Cargo.toml
@@ -8,6 +8,7 @@ slog = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 anyhow = { workspace = true }
+uuid = { workspace = true }
 
 [features]
 moose = []
--- a/drop-analytics/src/file_impl.rs
+++ b/drop-analytics/src/file_impl.rs
@@ -2,6 +2,7 @@ use std::{fs::File, io::Write, path::Path};
 
 use serde::{Deserialize, Serialize};
 use slog::Logger;
+use uuid;
 
 use crate::{FileInfo, TransferDirection, TransferInfo, MOOSE_STATUS_SUCCESS, MOOSE_VALUE_NONE};
 
@@ -78,7 +79,13 @@ impl FileImpl {
         };
         events.push(event);
 
-        File::create(&self.event_path)?.write_all(serde_json::to_string(&events)?.as_bytes())?;
+        let payload = serde_json::to_string(&events)?;
+
+        // create unique temp path
+        let temp_path = format!("{}.tmp.{}", self.event_path, uuid::Uuid::new_v4());
+
+        File::create(&temp_path)?.write_all(payload.as_bytes())?;
+        std::fs::rename(temp_path, &self.event_path)?;
 
         Ok(())
     }
--- a/drop-analytics/src/file_impl.rs
+++ b/drop-analytics/src/file_impl.rs
@@ -84,7 +84,7 @@ impl FileImpl {
         // create unique temp path
         let temp_path = format!("{}.tmp.{}", self.event_path, uuid::Uuid::new_v4());
 
-        File::create(&temp_path)?.write_all(payload.as_bytes())?;
+        std::fs::write(&temp_path, payload.as_bytes())?;
         std::fs::rename(temp_path, &self.event_path)?;
 
         Ok(())
--- a/drop-transfer/src/service.rs
+++ b/drop-transfer/src/service.rs
@@ -89,39 +89,13 @@ impl Service {
 
             ws::server::spawn(state.clone(), logger.clone(), stop.clone(), guard.clone())?;
 
-            let mut service = Self {
+            Ok(Self {
                 state,
                 stop,
                 waiter,
                 logger,
                 tasks: Default::default(),
-            };
-
-            let outgoing_transfers = {
-                let xfers = service.state.transfer_manager.outgoing.lock().await;
-                xfers
-                    .values()
-                    .map(|xstate| xstate.xfer.clone())
-                    .collect::<Vec<_>>()
-            };
-
-            for xfer in outgoing_transfers {
-                service.trigger_peer_outgoing(xfer);
-            }
-
-            let incoming_transfers = {
-                let xfers = service.state.transfer_manager.incoming.lock().await;
-                xfers
-                    .values()
-                    .map(|xstate| xstate.xfer.clone())
-                    .collect::<Vec<_>>()
-            };
-
-            for xfer in incoming_transfers {
-                service.trigger_peer_incoming(xfer);
-            }
-
-            Ok(service)
+            })
         };
 
         let res = task.await;
--- a/drop-transfer/src/service.rs
+++ b/drop-transfer/src/service.rs
@@ -189,38 +189,42 @@ impl Service {
 
     pub async fn set_peer_state(&mut self, addr: IpAddr, is_online: bool) {
         {
-            let outgoing_transfers_to_trigger = {
-                let xfers = self.state.transfer_manager.outgoing.lock().await;
+            if is_online {
+                let outgoing_transfers_to_trigger = {
+                    let xfers = self.state.transfer_manager.outgoing.lock().await;
 
-                xfers
-                    .values()
-                    .filter(|state| {
-                        let peer = state.xfer.peer();
-                        peer == addr && is_online
-                    })
-                    .map(|state| state.xfer.clone())
-                    .collect::<Vec<_>>()
-            };
+                    xfers
+                        .values()
+                        .filter(|state| {
+                            let peer = state.xfer.peer();
+                            peer == addr
+                        })
+                        .map(|state| state.xfer.clone())
+                        .collect::<Vec<_>>()
+                };
 
-            for xfer in outgoing_transfers_to_trigger {
-                self.trigger_peer_outgoing(xfer);
+                for xfer in outgoing_transfers_to_trigger {
+                    self.trigger_peer_outgoing(xfer);
+                }
             }
 
-            let incoming_transfers_to_trigger = {
-                let xfers = self.state.transfer_manager.incoming.lock().await;
+            if is_online {
+                let incoming_transfers_to_trigger = {
+                    let xfers = self.state.transfer_manager.incoming.lock().await;
 
-                xfers
-                    .values()
-                    .filter(|state| {
-                        let peer = state.xfer.peer();
-                        peer == addr && is_online
-                    })
-                    .map(|state| state.xfer.clone())
-                    .collect::<Vec<_>>()
-            };
+                    xfers
+                        .values()
+                        .filter(|state| {
+                            let peer = state.xfer.peer();
+                            peer == addr && is_online
+                        })
+                        .map(|state| state.xfer.clone())
+                        .collect::<Vec<_>>()
+                };
 
-            for xfer in incoming_transfers_to_trigger {
-                self.trigger_peer_incoming(xfer);
+                for xfer in incoming_transfers_to_trigger {
+                    self.trigger_peer_incoming(xfer);
+                }
             }
         }
     }
--- a/drop-transfer/src/service.rs
+++ b/drop-transfer/src/service.rs
@@ -106,7 +106,7 @@ impl Service {
             };
 
             for xfer in outgoing_transfers {
-                service.trigger_peer_outgoing(xfer).await;
+                service.trigger_peer_outgoing(xfer);
             }
 
             let incoming_transfers = {
@@ -118,7 +118,7 @@ impl Service {
             };
 
             for xfer in incoming_transfers {
-                service.trigger_peer_incoming(xfer).await;
+                service.trigger_peer_incoming(xfer);
             }
 
             Ok(service)
@@ -140,10 +140,9 @@ impl Service {
         &self.state.storage
     }
 
-    async fn trigger_peer_incoming(&mut self, xfer: Arc<crate::IncomingTransfer>) {
+    fn trigger_peer_incoming(&mut self, xfer: Arc<crate::IncomingTransfer>) {
         debug!(self.logger, "trigger_peer_incoming() called: {:?}", xfer);
 
-        let _peer = xfer.peer();
         let id = xfer.id();
 
         if let Some(handle) = self.tasks.get(&id) {
@@ -165,11 +164,10 @@ impl Service {
         self.tasks.insert(id, handle);
     }
 
-    async fn trigger_peer_outgoing(&mut self, xfer: Arc<crate::OutgoingTransfer>) {
+    fn trigger_peer_outgoing(&mut self, xfer: Arc<crate::OutgoingTransfer>) {
         debug!(self.logger, "trigger_peer_outgoing() called: {:?}", xfer);
 
         let id = xfer.id();
-        let _peer = xfer.peer();
         if let Some(handle) = self.tasks.get(&id) {
             if !handle.is_finished() {
                 return;
@@ -189,7 +187,7 @@ impl Service {
         self.tasks.insert(id, handle);
     }
 
-    pub async fn set_peer_state(&mut self, addr: &IpAddr, is_online: bool) {
+    pub async fn set_peer_state(&mut self, addr: IpAddr, is_online: bool) {
         {
             let outgoing_transfers_to_trigger = {
                 let xfers = self.state.transfer_manager.outgoing.lock().await;
@@ -198,14 +196,14 @@ impl Service {
                     .values()
                     .filter(|state| {
                         let peer = state.xfer.peer();
-                        peer == *addr && is_online
+                        peer == addr && is_online
                     })
                     .map(|state| state.xfer.clone())
                     .collect::<Vec<_>>()
             };
 
             for xfer in outgoing_transfers_to_trigger {
-                self.trigger_peer_outgoing(xfer).await;
+                self.trigger_peer_outgoing(xfer);
             }
 
             let incoming_transfers_to_trigger = {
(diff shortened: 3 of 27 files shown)


drop-analytics/src/file_impl.rs
```diff
@@ -78,7 +79,13 @@ impl FileImpl {
...
         events.push(event);
 
-        File::create(&self.event_path)?.write_all(serde_json::to_string(&events)?.as_bytes())?;
+        let payload = serde_json::to_string(&events)?;
+
+        // create unique temp path
+        let temp_path = format!("{}.tmp.{}", self.event_path, uuid::Uuid::new_v4());
+
+        File::create(&temp_path)?.write_all(payload.as_bytes())?;
+        std::fs::rename(temp_path, &self.event_path)?;
```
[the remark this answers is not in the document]
LukasPukenis: Sure thing! I've started to notice weird logs from this place about invalid JSON. When printing it it was really invalid JSON, no idea how it could be reproduced until it hit me that when using tests that are doing a lot of writes to the database, it might be that several of these function calls get interlaced. Because the file is written in chunks, some chunks would be interlaced resulting in invalid JSON.

The usual way for such cases is to write to a unique temp file and then move it to the final destination(fs move is an atomic operation). However, even though this fixes this exact annoyance it will still render the file non-deterministic in that case as some stuff will be lost. The proper way would be to have a mutex around the writer as well but  it requires much more changes

drop-transfer/src/service.rs
```diff
@@ -111,6 +143,91 @@ impl Service {
         &self.state.storage
     }
 
+    async fn trigger_peer_incoming(&self, xfer: Arc<crate::IncomingTransfer>) {
+        debug!(self.logger, "trigger_peer inc() called: {:?}", xfer);
```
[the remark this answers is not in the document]
LukasPukenis: agreed :)

drop-transfer/src/service.rs
```diff
@@ -85,14 +101,30 @@ impl Service {
...
+                }
+            }
+
+            {
+                let xfers = service.state.transfer_manager.incoming.lock().await;
+
+                for xstate in xfers.values() {
+                    service.trigger_peer_incoming(xstate.xfer.clone()).await;
+                }
+            }
```
Lipt0nas, line 125: Im thinking if it makes sense to trigger all all of the incoming and outgoing states on service start, in the situation that someone has a lot of transfers queued up might not be optimal? Could we not just instead rely on the apps to give us this information? We could for example require this peer state JSON in the config on startup to make sure the apps don't forget to give us the initial state

drop-analytics/Cargo.toml
```diff
@@ -8,6 +8,7 @@ slog = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 anyhow = { workspace = true }
+uuid = { workspace = true }
```
matszczygiel, line 11: It would be nice to include this dependency only when the `moose_file` feature is on

drop-analytics/src/file_impl.rs
```diff
@@ -78,7 +79,13 @@ impl FileImpl {
         };
         events.push(event);
 
-        File::create(&self.event_path)?.write_all(serde_json::to_string(&events)?.as_bytes())?;
+        let payload = serde_json::to_string(&events)?;
+
+        // create unique temp path
+        let temp_path = format!("{}.tmp.{}", self.event_path, uuid::Uuid::new_v4());
+
+        File::create(&temp_path)?.write_all(payload.as_bytes())?;
```
matszczygiel, line 87:
```suggestion
        std::fs::write(&temp_path, payload.as_bytes())?;
```

drop-analytics/src/file_impl.rs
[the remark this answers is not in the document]
matszczygiel: If that's the case then `Mutex` is a way to go IMHO. With you're solution it's possible to lose some events

drop-transfer/src/transfer.rs
```diff
@@ -72,6 +72,10 @@ impl<F: File> TransferData<F> {
         Self::new_with_uuid(peer, files, Uuid::new_v4(), config)
     }
 
+    pub(crate) fn peer(&self) -> IpAddr {
+        self.peer
+    }
+
```
matszczygiel, line 78: @LukasPukenis there is already this method in the `Transfer` trait which is implemented on `TransferData<F>`
```suggestion

```

drop-transfer/src/service.rs
```diff
@@ -36,14 +38,28 @@ pub(super) struct State {
...
+#[derive(Deserialize, Debug)]
+pub enum PeerState {
+    #[serde(rename = "online")]
+    Online,
+    #[serde(rename = "offline")]
+    Offline,
+}
+
+type TaskRegistry =
+    Arc<tokio::sync::Mutex<HashMap<(IpAddr, uuid::Uuid), tokio::task::JoinHandle<()>>>>;
```
matszczygiel, line 52: I don't think you need `IpAddr` where. It's unambiguously defined by the `Uuid` since transfers have only one peer (IP)

matszczygiel: Moreover, you never call `clone()` on it so `Arc` is not necessary here and adds unnecessary indirection

drop-transfer/src/ws/client/mod.rs
```diff
@@ -269,37 +260,14 @@ async fn make_request(
     Err(err)
 }
 
-async fn tcp_connect<Fut: Future<Output = bool>>(
-    local_ip: IpAddr,
-    remote_ip: IpAddr,
-    break_check: impl Fn() -> Fut,
-    logger: &Logger,
-) -> Option<TcpStream> {
+async fn tcp_connect(local_ip: IpAddr, remote_ip: IpAddr, logger: &Logger) -> Option<TcpStream> {
```
matszczygiel, line 263: Is this function worth keeping? It's so insignificant at this point that it would be less cumbersome to just inline it

drop-transfer/src/service.rs
matszczygiel: Going deeper you don't need `Mutex`, since `Service` is being passed as a mutable reference

norddrop/src/device/mod.rs
```diff
@@ -371,6 +373,32 @@ impl NordDropFFI {
...
+        };
+
+        debug!(self.logger, "Parsed set_peer_states(): {:?}", states);
+
+        let mut instance = self.instance.blocking_lock();
+        let instance = instance
+            .as_mut()
+            .ok_or(ffi::types::NORDDROP_RES_NOT_STARTED)?;
+
+        self.rt.block_on(instance.set_peers_state(states));
```
matszczygiel, line 397: I'm not sure if this is in line with other methods. I think we use `spawn()` not `block_on()` but I don't have a strong opinion on which one we should use. Just saying it's inconsistent

norddrop/src/ffi/mod.rs
```diff
@@ -675,6 +675,53 @@ pub unsafe extern "C" fn norddrop_new(
...
+            Ok(inst) => inst,
+            Err(poisoned) => poisoned.into_inner(),
+        };
+
+        let states = {
+            if states.is_null() {
+                return norddrop_result::NORDDROP_RES_INVALID_STRING;
+            }
+
+            ffi_try!(CStr::from_ptr(states).to_str())
```
matszczygiel, line 715:
```suggestion
            Str::from_ptr(states).to_str()?
```
There is an implementation `impl From<std::str::Utf8Error> for norddrop_result`

matszczygiel (commented):
In general, it looks OK with minor tweaks left. However, I cannot understand why we use such a complicated API for this. Isn't API like this
```c
norddrop_result set_peer_status(const char* peer, int is_online);
```
cleaner and easier to use and implement?
I don't see a point in batching statuses in a JSON structure

drop-analytics/src/file_impl.rs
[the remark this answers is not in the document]
LukasPukenis: I think the current variant is still an improvement(no invalid json is produced), we can implement the mutexed solution once this causes us problems

drop-transfer/src/service.rs
LukasPukenis: That is actually a fair point. I will try to remove it and see how it goes

norddrop/src/device/mod.rs
LukasPukenis: I've checked and we actually use `block_on` in other methods: `transfers_since`, `remove_transfer_file`, `new_transfer`, `start`, `stop`, `purge_transfers`, `purge_transfers_until`, so actually all of them :D

drop-transfer/src/ws/client/mod.rs
```diff
@@ -341,6 +309,7 @@ impl RunContext<'_> {
                 };
                 if let Err(err) = task.await {
                     error!(self.logger, "Failed to close socket on start: {err:?}");
+                // TODO: what does that mean
```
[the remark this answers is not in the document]
LukasPukenis: thanks for explaining

drop-transfer/src/service.rs
LukasPukenis: You're right about the UUID and peer. This simplified things

norddrop/src/device/mod.rs
matszczygiel: Oh, then it's all good :+1: . Sorry for not checking it myself

LukasPukenis: no worries at all

drop-transfer/src/service.rs
LukasPukenis: Mutex point is also truth, I simplified it

drop-transfer/src/service.rs
LukasPukenis: I've added it and surprisingly to me - no testcase broke because of it which is a bonus

Lipt0nas: nice!

drop-transfer/src/service.rs
```diff
@@ -111,6 +140,93 @@ impl Service {
...
+            xfer,
+            self.logger.clone(),
+            self.waiter.guard(),
+            self.stop.clone(),
+        );
+
+        self.tasks.insert(id, handle);
+    }
+
+    pub async fn set_peer_state(&mut self, addr: &IpAddr, is_online: bool) {
```
matszczygiel, line 192:
```suggestion
    pub async fn set_peer_state(&mut self, addr: IpAddr, is_online: bool) {
```
`IpAddr` is copyable

drop-transfer/src/service.rs
```diff
@@ -111,6 +140,93 @@ impl Service {
...
+        {
+            let outgoing_transfers_to_trigger = {
+                let xfers = self.state.transfer_manager.outgoing.lock().await;
+
+                xfers
+                    .values()
+                    .filter(|state| {
+                        let peer = state.xfer.peer();
+                        peer == *addr && is_online
```
matszczygiel, line 201: The `is_online` check should go above, since if it is `false` it renders the whole iteration unnecessary

drop-transfer/src/service.rs
```diff
@@ -111,6 +140,93 @@ impl Service {
...
+                    })
+                    .map(|state| state.xfer.clone())
+                    .collect::<Vec<_>>()
```
matszczygiel, line 204: Minor: I bet this could've been implemented without allocating new `Vec` with more fine-grained control over `Service` struct fields.

drop-transfer/src/service.rs
```diff
@@ -111,6 +140,93 @@ impl Service {
...
+        );
+
+        self.tasks.insert(id, handle);
+    }
+
+    async fn trigger_peer_outgoing(&mut self, xfer: Arc<crate::OutgoingTransfer>) {
+        debug!(self.logger, "trigger_peer_outgoing() called: {:?}", xfer);
+
+        let id = xfer.id();
+        let _peer = xfer.peer();
```
matszczygiel, line 172:
```suggestion
```

drop-transfer/src/service.rs
```diff
@@ -111,6 +140,93 @@ impl Service {
         &self.state.storage
     }
 
+    async fn trigger_peer_incoming(&mut self, xfer: Arc<crate::IncomingTransfer>) {
+        debug!(self.logger, "trigger_peer_incoming() called: {:?}", xfer);
+
+        let _peer = xfer.peer();
```
matszczygiel, line 146:
```suggestion
```

matszczygiel (changes requested):

drop-transfer/src/service.rs
LukasPukenis: That is true, though it makes look the code awkward

LukasPukenis:
@matszczygiel @Lipt0nas:
- I've squashed the history into a few final commits
- Simplified the API for singular peer
- Removed auto initialization on drop start thus relying on explicit call to the new API

drop-transfer/src/service.rs
matszczygiel: IMHO performance and battery life are at stake here, since this iteration might not be cheap if there are a lot of transfers

LukasPukenis: fixed

drop-transfer/src/service.rs
LukasPukenis: No idea about this to be honest

drop-transfer/src/service.rs
```diff
@@ -85,14 +89,39 @@ impl Service {
...
+                let xfers = service.state.transfer_manager.incoming.lock().await;
+                xfers
+                    .values()
+                    .map(|xstate| xstate.xfer.clone())
+                    .collect::<Vec<_>>()
+            };
+
+            for xfer in incoming_transfers {
+                service.trigger_peer_incoming(xfer);
+            }
```
Lipt0nas, line 122: I thought that we agreed to remove the triggering of of transfers on service start

LukasPukenis: No idea how it slipped back in(or possibly was never removed). Pushed the change just now

drop-analytics/src/file_impl.rs
matszczygiel: It's a minor thing but it still has not changed. If you want to ignore it, go ahead

drop-transfer/src/service.rs
```diff
@@ -111,6 +114,95 @@ impl Service {
...
+        debug!(self.logger, "trigger_peer_incoming() called: {:?}", xfer);
+
+        let id = xfer.id();
+
+        if let Some(handle) = self.tasks.get(&id) {
+            if !handle.is_finished() {
+                return;
+            } else {
+                self.tasks.remove(&id);
+            }
```
matszczygiel, line 127:
```suggestion
            }
```
It's a minor thing but the `insert()` below will remove it anyway in a more efficient way

drop-transfer/src/service.rs
```diff
@@ -111,6 +114,95 @@ impl Service {
...
+    fn trigger_peer_outgoing(&mut self, xfer: Arc<crate::OutgoingTransfer>) {
+        debug!(self.logger, "trigger_peer_outgoing() called: {:?}", xfer);
+
+        let id = xfer.id();
+        if let Some(handle) = self.tasks.get(&id) {
+            if !handle.is_finished() {
+                return;
+            } else {
+                self.tasks.remove(&id);
+            }
```
matszczygiel, line 150: Same here

drop-transfer/src/service.rs
```diff
@@ -111,6 +114,95 @@ impl Service {
...
+                        .map(|state| state.xfer.clone())
+                        .collect::<Vec<_>>()
+                };
+
+                for xfer in outgoing_transfers_to_trigger {
+                    self.trigger_peer_outgoing(xfer);
+                }
+            }
+
+            if is_online {
```
matszczygiel, line 185:
```suggestion


```

drop-transfer/src/service.rs
```diff
@@ -111,6 +114,95 @@ impl Service {
...
+                let incoming_transfers_to_trigger = {
+                    let xfers = self.state.transfer_manager.incoming.lock().await;
+
+                    xfers
+                        .values()
+                        .filter(|state| {
+                            let peer = state.xfer.peer();
+                            peer == addr && is_online
```
matszczygiel, line 193:
```suggestion
                            peer == addr
```

drop-analytics/Cargo.toml
LukasPukenis: too many comments, gonna tackle those of higher priority

drop-analytics/src/file_impl.rs
LukasPukenis: I wonder what difference does it make. It should be in faster, in theory, without creating an instance but still wonder

drop-transfer/src/service.rs
LukasPukenis: updated it

drop-analytics/src/file_impl.rs
LukasPukenis: updated it

matszczygiel: Probably there is no difference. Just more direct API

matszczygiel (approved):

PR #232, opened 2023-09-29
Fix: https://github.com/NordSecurity/libdrop/security/dependabot/7

LukasPukenis:
Basically it's just tungstenite and warp version bumps


matszczygiel (approved):
:+1:

PR #233, opened 2023-10-02
Guard `norddrop_start()` from multiple calls

matszczygiel:
This basically ensures that all the preconditions (including if it's already started) are met before the actual initialization.
I've also added a test case just to be sure.


Lipt0nas (approved):
+1 for including a test case as well

PR #235, opened 2023-10-03
Rust 1.72.1

Lipt0nas:
It's debatable if we should just use `#[allow]` everywhere for the unused variables, as we did before, but figured more logging can always help

### DIFF
--- a/norddrop/src/ffi/mod.rs
+++ b/norddrop/src/ffi/mod.rs
@@ -119,6 +119,7 @@ pub unsafe extern "C" fn norddrop_new_transfer(
 #[no_mangle]
 pub unsafe extern "C" fn norddrop_destroy(dev: *mut norddrop) {
     if !dev.is_null() {
+        #[allow(clippy::let_underscore_must_use)]
         let _ = Box::from_raw(dev);
     }
 }
@@ -773,6 +774,7 @@ impl slog::Drain for norddrop_logger_cb {
         let kv = record.kv();
 
         let mut serializer = KeyValueSerializer::new(record);
+        #[allow(clippy::let_underscore_must_use)]
         let _ = kv.serialize(record, &mut serializer);
 
         if let Ok(cstr) = CString::new(serializer.msg()) {
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.64.0
+          toolchain: 1.72.1
           default: true        
       - name: docker sdk
         run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,14 +6,14 @@ on:
     branches: [dev, main]
 jobs:
   lint-rust:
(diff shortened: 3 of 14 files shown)


LukasPukenis (approved):

matszczygiel:
@Lipt0nas I disagree that the lint is needed. It's off by default in clippy for a reason.
The `let _ =` is intentional in our codebase and I'd argue it should be left as is since in some cases an error is not actually an error.

Lipt0nas:
This lint was actually enabled for a very long time now, wonder why it only appeared now

I would kind of agree that it's mostly not needed, as it generally introduces more trouble than it helps

@LukasPukenis What is your opinion on the lint?

matszczygiel:
By the way, should we merge this into `main` instead of `dev`? Or possibly both?

LukasPukenis (approved):
First of all, I think it's a nice change to handle return values with logging instead of silently failing. This is a really nice pattern and should pay off when sifting through app logs.

norddrop/src/ffi/mod.rs
```diff
@@ -119,6 +119,7 @@ pub unsafe extern "C" fn norddrop_new_transfer(
 #[no_mangle]
 pub unsafe extern "C" fn norddrop_destroy(dev: *mut norddrop) {
     if !dev.is_null() {
+        #[allow(clippy::let_underscore_must_use)]
         let _ = Box::from_raw(dev);
     }
```
LukasPukenis, line 124: we could do `let inst = Box::from_raw(dev);` this way the lint comment would not be needed, however it's one occasion, it doesn't really matter

Lipt0nas: then we'd have to use `allow unused variables` instead of this to make the lint stage pass :D

Lipt0nas:
@matszczygiel id say makes sense to merge this into dev, as this is the branch we're actively working on, not sure if merging to main instantly is needed, as our next release should contain the changes present in dev right now

matszczygiel:
@LukasPukenis I somewhat disagree with your take. From the rust reference https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute it's idiomatic to use `let _ =` when you want to discard the value (read the bottom box).
For example, the false positive you mentioned with `Box::from_raw()`.
BTW @Lipt0nas you may write `drop(Box::from_raw())` to avoid allowing lint

LukasPukenis:
@matszczygiel I think it's bikeshedding at this point regarding the Box clippy part.

@Lipt0nas merging to `dev` allows us to make a release with @matszczygiel fixes as well and call it v5.3 or v5.2.1 so I would vote for dev now as I prefer smaller releases and have v6 only with new API

matszczygiel (approved):

matszczygiel:
Sorry, for the confusion but have we agreed on merging this to `main` or `dev`? I think it's better to merge it to `main` since it will be merged back into `dev` anyway with the next release

LukasPukenis:
@Lipt0nas `main` it is

matszczygiel:
@Lipt0nas I think this one should be closed in favor of https://github.com/NordSecurity/libdrop/pull/241

Lipt0nas:
yeah, this no longer needs to be here

PR #236, opened 2023-10-03
Attempt to fix missing `FileDownloaded` event on rapidly disconnectin…

matszczygiel:
So this branch contains fixes from the `dev` branch and the fix to issues with turning meshnet on and off.

The fix is contained in the last two commits, so it's worth reviewing.
The older ones are simply copied from the `dev` branch and were reviewed before, so there is no point in reading that

### DIFF
--- a/drop-transfer/src/ws/events.rs
+++ b/drop-transfer/src/ws/events.rs
@@ -177,7 +177,7 @@ impl FileEventTx<IncomingTransfer> {
     }
 
     pub async fn success(&self, final_path: impl Into<PathBuf>) {
-        self.stop(
+        self.force_stop(
             crate::Event::FileDownloadSuccess(
                 self.xfer.clone(),
                 crate::event::DownloadSuccess {
--- a/drop-transfer/src/ws/server/mod.rs
+++ b/drop-transfer/src/ws/server/mod.rs
@@ -745,6 +745,7 @@ impl FileXferTask {
         Ok(dst)
     }
 
+    #[allow(clippy::too_many_arguments)]
     async fn run(
         mut self,
         state: Arc<State>,
@@ -753,6 +754,7 @@ impl FileXferTask {
         mut stream: UnboundedReceiver<Vec<u8>>,
         req_send: mpsc::UnboundedSender<ServerReq>,
         logger: Logger,
+        guard: AliveGuard,
     ) {
         let init_res = match downloader.init(&self).await {
             Ok(init) => init,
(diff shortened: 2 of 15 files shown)


drop-transfer/src/ws/events.rs
```diff
@@ -177,7 +177,7 @@ impl FileEventTx<IncomingTransfer> {
     }
 
     pub async fn success(&self, final_path: impl Into<PathBuf>) {
-        self.stop(
+        self.force_stop(
```
LukasPukenis, line 180: Can you explain this change? I've also taken a look into `stop` and `force_stop` and they seem to be identical? It's difficult to understand what this commit has to do with the commit message to be honest

drop-transfer/src/ws/server/mod.rs
```diff
@@ -938,9 +958,9 @@ impl<'a> FileStreamCtx<'a> {
             } = self;
 
             jobs.spawn(async move {
-                let _guard = guard;
+                let _gurad = guard.clone();
 
-                task.run(state, events, downloader, stream, req_send, logger)
+                task.run(state, events, downloader, stream, req_send, logger, guard)
```
LukasPukenis, line 963: why do we clone the guard and not use it? also it contains a typo

matszczygiel: The copy is passed in because the function `run()` can spawn its own tasks, which we want to have control over when they finish.
I'll fix the typo :+1:

drop-transfer/src/ws/events.rs
matszczygiel: `force_stop()` does not require the file to be started first during the same connection. Look at the `success()` call for the `OutgoingTransfer`.
`stop()` silently ignores the event if the file was not started, it's meant to be used with less critical events like `Progress` for example.

drop-transfer/src/ws/server/mod.rs
matszczygiel: Typo fixed

LukasPukenis: but I mean why do we even need the `_guard` if we don't use it - do we gain something from making a copy of guard and dropping it almost immediately?

drop-transfer/src/ws/events.rs
LukasPukenis: My bad, didn't notice that we early return in the `stop` case.

drop-transfer/src/ws/server/mod.rs
matszczygiel: It's not dropped immediately. `run()` function is async and the `_guard` is dropped only after `run()` finishes the execution

LukasPukenis: what would happen if that line would be deleted?

matszczygiel: Which one?
If you meant this one `let _gurad = guard.clone();` then there is a possibility of shutting down the tokio runtime in `norddrop_close()` before the task finishes. We've seen this before, it leads to strange behaviors like random panics and so on.
In fact, the guards were introduced just to cope with this problem

matszczygiel: The `norddrop_close()` waits for all the guards to be dropped and only then return to the caller

LukasPukenis: thanks for explaining

LukasPukenis (approved):

PR #237, opened 2023-10-03
Lukasp/fix testcases after noddrop set peer

LukasPukenis:
WaitForAnotherPeer() was broken for ipv6
Few other tests were broken because there was not an appropriate WaitForAnotherPeer

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -2192,6 +2192,7 @@ scenarios = [
                         "DROP_PEER_STIMPY",
                         "/tmp/testfile-big",
                     ),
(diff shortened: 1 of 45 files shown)


test/scenarios.py
```diff
@@ -3517,6 +3517,7 @@
         {
             "DROP_PEER_REN": ActionList(
                 [
+                    action.ConfigureNetwork(latency="200ms"),
```
matszczygiel, line 3520: Why so much latency here?

LukasPukenis: Actually no reason for it to be there as testrunner queues the events, removed this one :)

matszczygiel (approved):

PR #238, opened 2023-10-04
Improve temporaries handling

matszczygiel:
This PR contains the following changes:
* Ensure that all outstanding tasks are finished before removing temporaries while finalizing
* More logging added
* Try to remove temporaries right away when rejecting

### DIFF
--- a/drop-storage/src/lib.rs
+++ b/drop-storage/src/lib.rs
@@ -1340,6 +1340,42 @@ impl Storage {
         }
     }
 
+    pub async fn fetch_base_dirs_for_file(&self, transfer_id: Uuid, file_id: &str) -> Vec<String> {
+        let tid = transfer_id.to_string();
+
+        trace!(
+            self.logger,
+            "Fetching temporary file locations";
+            "transfer_id" => &tid
+        );
+
+        let task = async {
+            let conn = self.conn.lock().await;
+
+            let out = conn
+                .prepare(
+                    r#"
+                SELECT DISTINCT base_dir
+                FROM incoming_paths ip
+                INNER JOIN incoming_path_pending_states ipss ON ip.id = ipss.path_id 
+                WHERE transfer_id = ?1 AND path_hash = ?2
+                "#,
+                )?
+                .query_map(params![tid, file_id], |row| row.get("base_dir"))?
+                .collect::<QueryResult<_>>()?;
+
+            Ok::<Vec<_>, Error>(out)
+        };
+
+        match task.await {
+            Ok(res) => res,
+            Err(e) => {
+                error!(self.logger, "Failed to fetch temporary file locations for {file_id}"; "error" => %e);
+                vec![]
+            }
+        }
+    }
+
     pub async fn cleanup_garbage_transfers(&self) -> usize {
         trace!(self.logger, "Removing garbage transfers");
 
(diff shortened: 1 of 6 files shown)


test/scenarios.py
```diff
@@ -8605,4 +8605,218 @@
...
+                    action.CancelTransferRequest(0),
+                    action.ExpectCancel([0], False),
+                    action.Stop(),
+                ]
+            ),
+        },
+    ),
+    Scenario(
+        "scenario45-3",
+        "Assert the temporary files are removed right after the rejection from the sender and the receiver is offline",
```
LukasPukenis, line 8756: The sender goes offline, not the receiver in this testcase.

drop-storage/src/lib.rs
```diff
@@ -1340,6 +1340,46 @@ impl Storage {
         }
     }
 
+    pub async fn fetch_temp_locations_for_file(
```
LukasPukenis, line 1343: It is a bit misleading, as this function fetches the base directories for downloads and has no knowledge of temporary files, it could be named `fetch_base_directories_for_file()` for example.

Question - do we expect multiple values? I assume it's just a failsafe just in case something went bad to remove the temp files from the previous downloads?

matszczygiel: Thanks :+1: . I will change the name. I'm not sure, but shouldn't be possible. However, it's safer to fetch all of them if present in the table

test/scenarios.py
matszczygiel: Oh yeah. I must have forgotten to update the description while making changes

matszczygiel: Fixed

drop-storage/src/lib.rs
matszczygiel: Function name changed

drop-transfer/src/ws/server/mod.rs
```diff
@@ -479,18 +479,21 @@ impl RunContext<'_> {
             anyhow::Ok(())
         };
 
-        if let Err(err) = task.await {
+        let result = task.await;
+        info!(self.logger, "Connection loop finished");
+
+        jobs.shutdown().await;
+
+        if let Err(err) = result {
```
LukasPukenis, line 488: Can we use `jobs` to spawn the `task` and just have `jobs.shutdown().await` alone instead of awaiting for the task explicitly? I am not sure why we need both and it feels like it could produce a race.

drop-storage/src/lib.rs
LukasPukenis: I agree that it's safer, lets leave it as it is :)

drop-transfer/src/ws/server/mod.rs
matszczygiel: Um, no. We need to make use of a `result` of that `task`. The second thing is that the task is local to this function. It is not spawned. The third thing is that we actually want the task to be finished before we close any outstanding tasks.

I think even the compiler would prevent that because of the borrowing rules

LukasPukenis: I was just curious, it seems that we use this `jobs` as a task manager of some sorts, so assumed we can even move the `task` into there. I wonder what will happen if the peer goes offline for example - is the task gonna fail earlier thanone of the the spawned jobs?

matszczygiel: I'm afraid we can't move the task there. The task is basically the one that uses the `jobs` manager, so it would mean that the manager manages itself :shrug:

LukasPukenis (dismissed):

LukasPukenis (approved):

PR #241, opened 2023-10-04
Rust 1.72.1 - main edition

Lipt0nas:
Or is it actually `dev`?

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.64.0
+          toolchain: 1.72.1
           default: true        
       - name: docker sdk
         run: |
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,14 +6,14 @@ on:
     branches: [dev, main]
 jobs:
   lint-rust:
-    container: rust:1.64-bullseye
+    container: rust:1.72.1-bullseye
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.64.0
+          toolchain: 1.72.1
           default: true
           components: clippy
       - name: Install nightly Rust toolchain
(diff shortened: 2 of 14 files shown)


matszczygiel:
@Lipt0nas it depends if we want to include it in 5.3 or not. I'd say let's do it in 5.3 thus merge it into the `main`

Lipt0nas:
Yeah, i guess this was the consensus, but it took A LOT of back and forths to get here :D

matszczygiel:
@Lipt0nas I have one thing though. Since the errors were ignored before, I think the `error!` logging severity would clutter logs a lot with errors. Of course, I'm against login at all in most of the cases but since the work is already done, let's change it to `warn!` maybe?

Lipt0nas:
@matszczygiel Almost all of the errors here are related to sending something trough a channel, alongside a few calls in Transfer Manager and such. The way i see it, these potential errors were ignored because they essentially should never happen, so it would make sense to log these in a way that should make the noticeable, were they to happen. Or am i wrong on their frequency?

matszczygiel:
Not exactly, these channels notify the connection logic to do some stuff. They will fail if the peer is offline, which since libdrop 5.0.0 is a normal thing

Lipt0nas:
Ah then i though about it in the wrong way, thanks for clarifying, ill change the logs to `warn`, and leave the `fs` and `tokio` calls as `error`

matszczygiel:
@Lipt0nas BTW the tokio join handle can be safely `unwrap()`ed here https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html

matszczygiel:
I meant that the error here is either because the task was canceled or panicked.
Cancelation is not an option because we await it directly, so no way to call `JoinHandle::cancel()` on it.
If panic occurs (well it shouldn't provided we have a severe bug in the code) then unwrapping it will propagate it further and since we're still in an async task, nothing would crash.

However, it seems to me that the best option is to ignore the result altogether as it was before but I don't want to push it further. It's your call what to do :+1:

matszczygiel (approved):

Lipt0nas:
So there is VERY unlikely scenario that it could panic if something was extremely wrong in the code, i think in this case it wouldn't hurt to log it rather than ignoring and propagating further also doesn't seem effective, given that this as unrecoverable as an error could get in this case

PR #245, opened 2023-10-06
set_peer_state -> refresh_network

LukasPukenis:
simplify code by having umbrella API for network change.

Along came:
- several retries upon the API call as it might be delayed
- fixed infinite check loop
- extra testcase

Adding retries simplified the logic where we don't need to store "refresh was called but the task was busy".

**TLDR:** Inherently this case makes the code behave as in current `main`, except instead of infinite retries it waits for the API to trigger it and gives out several tries.

### DIFF
--- a/drop-transfer/src/check.rs
+++ b/drop-transfer/src/check.rs
@@ -38,7 +38,6 @@ pub(crate) fn spawn(
                     break;
                 }
 
-                #[allow(clippy::let_underscore_must_use)]
                 let _ = refresh_trigger.changed().await;
             }
         };
--- a/drop-transfer/src/check.rs
+++ b/drop-transfer/src/check.rs
@@ -1,4 +1,4 @@
-use std::{net::SocketAddr, ops::ControlFlow, sync::Arc};
+use std::{net::SocketAddr, ops::ControlFlow, sync::Arc, time::Duration};
 
 use hyper::StatusCode;
 use slog::{debug, info, warn, Logger};
@@ -7,51 +7,71 @@ use tokio_util::sync::CancellationToken;
 use crate::{protocol, service::State, tasks::AliveGuard, IncomingTransfer, Transfer};
 
 pub(crate) fn spawn(
+    mut refresh_trigger: tokio::sync::watch::Receiver<()>,
     state: Arc<State>,
     xfer: Arc<IncomingTransfer>,
     logger: Logger,
     guard: AliveGuard,
     stop: CancellationToken,
-) -> tokio::task::JoinHandle<()> {
+) {
     let id = xfer.id();
-    let job = run(state, xfer, logger.clone());
 
     tokio::spawn(async move {
         let _guard = guard;
+
+        let task = async {
+            loop {
+                let cf = run(&state, &xfer, &logger).await;
+
+                if cf.is_break() {
+                    info!(logger, "Transfer {} is gone. Clearing", xfer.id());
+
+                    match state.transfer_manager.incoming_remove(xfer.id()).await {
+                        Err(err) => {
+                            warn!(logger, "Failed to clear incoming transfer: {err:?}");
+                        }
+                        Ok(false) => state
+                            .emit_event(crate::Event::IncomingTransferCanceled(xfer.clone(), true)),
+                        _ => (),
+                    }
+
+                    break;
+                }
+
+                #[allow(clippy::let_underscore_must_use)]
+                let _ = refresh_trigger.changed().await;
+            }
+        };
 
         tokio::select! {
             biased;
 
             _ = stop.cancelled() => {
-                debug!(logger, "Check job stop for transfer: {id}");
+                debug!(&logger, "stop checking job for: {}", id);
             },
-            _ = job => (),
+            _ = task => ()
         }
-    })
+    });
 }
 
-async fn run(state: Arc<State>, xfer: Arc<IncomingTransfer>, logger: Logger) {
-    loop {
-        tokio::time::sleep(drop_config::ALIVE_CHECK_INTERVAL).await;
+async fn run(state: &State, xfer: &Arc<IncomingTransfer>, logger: &Logger) -> ControlFlow<()> {
+    for delay in std::iter::once(Duration::from_secs(0)).chain(drop_config::RETRY_INTERVALS) {
+        debug!(
+            logger,
+            "Incoming transfer job started for {}, will sleep for {:?}",
+            xfer.id(),
+            delay
+        );
+        tokio::time::sleep(delay).await;
 
         if !state.transfer_manager.is_incoming_alive(xfer.id()).await {
-            break;
+            return ControlFlow::Break(());
         }
 
-        if make_request(&state, &xfer, &logger).await.is_break() {
-            break;
-        }
+        make_request(state, xfer, logger).await?;
     }
 
-    info!(logger, "Transfer {} is gone. Clearing", xfer.id());
-
-    match state.transfer_manager.incoming_remove(xfer.id()).await {
-        Err(err) => {
-            warn!(logger, "Failed to clear incoming transfer: {err:?}");
-        }
-        Ok(false) => state.emit_event(crate::Event::IncomingTransferCanceled(xfer.clone(), true)),
-        _ => (),
-    }
+    ControlFlow::Continue(())
 }
 
 async fn make_request(state: &State, xfer: &IncomingTransfer, logger: &Logger) -> ControlFlow<()> {
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -8,15 +8,19 @@ use std::{
 use anyhow::Context;
 use drop_config::DropConfig;
 use drop_storage::{sync, types::OutgoingFileToRetry, Storage};
-use slog::{debug, error, info, warn, Logger};
+use slog::{debug, error, info, trace, warn, Logger};
 use tokio::sync::{mpsc::UnboundedSender, Mutex};
+use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
 use crate::{
+    check,
     file::FileSubPath,
     service::State,
+    tasks::AliveGuard,
     transfer::{IncomingTransfer, OutgoingTransfer},
     ws::{
+        self,
         client::ClientReq,
         server::{FileXferTask, ServerReq},
         FileEventTx, FileEventTxFactory, IncomingFileEventTx, OutgoingFileEventTx,
@@ -656,6 +660,7 @@ impl TransferManager {
     }
 
     pub async fn outgoing_remove(&self, transfer_id: Uuid) -> crate::Result<()> {
+        debug!(self.logger, "Removing outgoing transfer: {}", transfer_id);
         let mut lock = self.outgoing.lock().await;
         if !lock.contains_key(&transfer_id) {
             return Err(crate::Error::BadTransfer);
@@ -677,6 +682,7 @@ impl TransferManager {
     }
 
     pub async fn outgoing_disconnect(&self, transfer_id: Uuid) -> crate::Result<()> {
+        trace!(self.logger, "outgoing_disconnect: {}", transfer_id);
         let mut lock = self.outgoing.lock().await;
         let _ = lock
             .get_mut(&transfer_id)
@@ -996,6 +1002,46 @@ pub(crate) async fn restore_transfers_state(state: &Arc<State>, logger: &Logger)
 
     let outgoing = restore_outgoing(state, logger).await;
     *state.transfer_manager.outgoing.lock().await = outgoing;
+}
+
+pub(crate) async fn resume(
+    refresh_trigger: &tokio::sync::watch::Receiver<()>,
+    state: &Arc<State>,
+    logger: &Logger,
+    guard: &AliveGuard,
+    stop: &CancellationToken,
+) {
+    {
+        let xfers = state.transfer_manager.outgoing.lock().await;
+
+        for xstate in xfers.values() {
+            let trig = refresh_trigger.clone();
+            ws::client::spawn(
+                trig,
+                state.clone(),
+                xstate.xfer.clone(),
+                logger.clone(),
+                guard.clone(),
+                stop.clone(),
+            );
+        }
+    }
+
+    {
+        let xfers = state.transfer_manager.incoming.lock().await;
+
+        for xstate in xfers.values() {
+            let trig = refresh_trigger.clone();
+            check::spawn(
+                trig,
+                state.clone(),
+                xstate.xfer.clone(),
+                logger.clone(),
+                guard.clone(),
+                stop.clone(),
+            );
+        }
+    }
 }
 
 async fn restore_incoming(
--- a/drop-transfer/src/service.rs
+++ b/drop-transfer/src/service.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::HashMap,
     fs,
     net::IpAddr,
     path::{Component, Path},
@@ -11,18 +10,19 @@ use drop_analytics::{InitEventData, Moose, TransferStateEventData};
 use drop_config::DropConfig;
 use drop_core::Status;
 use drop_storage::Storage;
-use slog::{debug, Logger};
+use slog::{debug, trace, Logger};
 use tokio::sync::{mpsc, Semaphore};
 use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
 use crate::{
-    auth, check,
+    auth,
     error::ResultExt,
     manager,
     tasks::AliveWaiter,
+    transfer::Transfer,
     ws::{self, FileEventTxFactory},
-    Error, Event, FileId, Transfer, TransferManager,
+    Error, Event, FileId, TransferManager,
 };
 
 pub(super) struct State {
@@ -38,8 +38,6 @@ pub(super) struct State {
     pub fdresolv: Option<Arc<crate::file::FdResolver>>,
 }
 
-type TaskRegistry = HashMap<uuid::Uuid, tokio::task::JoinHandle<()>>;
-
 impl State {
     pub fn emit_event(&self, event: crate::Event) {
         self.event_tx
@@ -54,7 +52,7 @@ pub struct Service {
     waiter: AliveWaiter,
     pub(super) logger: Logger,
 
-    tasks: TaskRegistry,
+    refresh_trigger: tokio::sync::watch::Sender<()>,
 }
 
 impl Service {
@@ -97,14 +95,23 @@ impl Service {
 
             manager::restore_transfers_state(&state, &logger).await;
 
-            ws::server::spawn(state.clone(), logger.clone(), stop.clone(), guard.clone())?;
+            let refresh_trigger = tokio::sync::watch::channel(()).0;
+            ws::server::spawn(
+                refresh_trigger.subscribe(),
+                state.clone(),
+                logger.clone(),
+                stop.clone(),
+                guard.clone(),
+            )?;
+
+            manager::resume(&refresh_trigger.subscribe(), &state, &logger, &guard, &stop).await;
 
             Ok(Self {
+                refresh_trigger,
                 state,
                 stop,
                 waiter,
-                logger,
-                tasks: Default::default(),
+                logger: logger.clone(),
             })
         };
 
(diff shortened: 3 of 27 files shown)


matszczygiel (commented):
To be honest the implementation feels very sloppy. I would like to see some improvements on the topic I mentioned in the comments

drop-transfer/src/check.rs
```diff
@@ -31,31 +31,43 @@ pub(crate) fn spawn(
 }
 
 async fn run(state: Arc<State>, xfer: Arc<IncomingTransfer>, logger: Logger) {
-    loop {
-        tokio::time::sleep(drop_config::ALIVE_CHECK_INTERVAL).await;
-
+    let mut remove_xfer = false;
```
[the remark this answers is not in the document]
LukasPukenis: Actually I've tried several iterations, what I wanted to do is `let remove_xfer = loop {` or `let remove_xfer = for delay in ...` but those are invalid, so resorted to Python

drop-transfer/src/service.rs
```diff
@@ -154,45 +166,70 @@ impl Service {
             self.stop.clone(),
         );
 
-        self.tasks.insert(id, handle);
+        guard.insert(id, handle);
     }
 
-    pub async fn set_peer_state(&mut self, addr: IpAddr, is_online: bool) {
+    async fn cleanup_tasks(&mut self) {
```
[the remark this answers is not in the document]
LukasPukenis: You need to comment more about what's wrong with this approach and what are the downsides of it. In my book its a clear improvement over the existing(leaky) implementation

drop-transfer/src/service.rs
```diff
@@ -45,7 +45,7 @@ pub struct Service {
     waiter: AliveWaiter,
     pub(super) logger: Logger,
 
-    tasks: TaskRegistry,
+    tasks: Arc<tokio::sync::Mutex<TaskRegistry>>,
```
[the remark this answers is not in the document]
LukasPukenis: Because this one is very simple and works just as well. If you see clear benefits of channeled approach in this case, feel free to share those

drop-transfer/src/check.rs
```diff
@@ -31,31 +31,43 @@ pub(crate) fn spawn(
...
         }
+
+        debug!(
+            logger,
+            "Client job will restart in {:?} seconds for {}",
+            delay,
+            xfer.id()
+        );
+
+        tokio::time::sleep(delay).await;
```
Lipt0nas, line 53: Doesn't sleeping at the end of the loop in this case mean that the last duration is effectively useless?

LukasPukenis: :+1: lemme fix it

drop-transfer/src/ws/server/mod.rs
```diff
@@ -166,14 +167,25 @@ pub(crate) fn spawn(
                     let stop = stop.clone();
                     let logger = logger.clone();
                     let nonces = nonces.clone();
+                    // TODO: why does this help
+                    let refresh_trigger = refresh_trigger.clone();
```
matszczygiel, line 171: Because you have `async move` down the line, so it moves the copy. Without it, you try to move multiple times the same variable

drop-transfer/src/check.rs
```diff
@@ -7,55 +7,84 @@ use tokio_util::sync::CancellationToken;
...
+                    break;
+                },
+                _ = job => {
+                    debug!(logger, "Check job done: {id}");
+                    break;
+                },
+                _ = refresh_trigger.changed() => {
+                    debug!(logger, "Check job refresh: {id}");
+                }
+            }
```
matszczygiel, line 39: I think this is incorrect. You're running the job while at the same time waiting for the channel.
This should be waiting and _then_ running the lob

matszczygiel: @LukasPukenis I think that's the root of the bug you found. The job is running and is suddenly killed when the channel receives a thing, thus breaking the connection abruptly.

drop-transfer/src/service.rs
```diff
@@ -114,85 +120,16 @@ impl Service {
...
-                };
-
-                for xfer in incoming_transfers_to_trigger {
-                    self.trigger_peer_incoming(xfer);
-                }
+            Err(err) => {
+                error!(
+                    self.logger,
+                    "Could not send a refresh trigger(probably channel closed): {:?}", err
+                )
```
matszczygiel, line 132: I think you can simply ignore the error here:
```rust
let _ = self.refresh_trigger.send(());
```
The error happens when there are no receivers alive. Which is not an error in our case

drop-transfer/src/ws/client/mod.rs
```diff
@@ -60,46 +60,79 @@ struct RunContext<'a> {
...
+                    debug!(logger, "Client job stop: {id}");
+                    break;
+                },
+                _ = job => {
+                    debug!(logger, "Client job done: {id}");
+                    break;
+                },
+                _ = refresh_trigger_clone.changed() => {
+                    debug!(logger, "Client job refresh: {id}");
+                }
```
matszczygiel, line 96: The same problem as in the server is here

drop-transfer/src/ws/client/mod.rs
```diff
@@ -60,46 +60,79 @@ struct RunContext<'a> {
...
+        if status.is_break() {
+            debug!(logger, "connection status is irrecoverable");
+            if let Err(err) = state.transfer_manager.outgoing_remove(xfer.id()).await {
+                warn!(
+                    logger,
+                    "Failed to clear sync state for {}: {err}",
+                    xfer.id()
+                );
+            }
+            return;
```
matszczygiel, line 123: It's the same as in the server. The breakage must be bubbled up

drop-transfer/src/ws/server/v4.rs
```diff
@@ -87,7 +87,7 @@ impl<'a> handler::HandlerInit for HandlerInit<'a> {
             .await
             .context("Failed to receive transfer request")?;
 
-        let msg = msg.to_str().ok().context("Expected JOSN message")?;
+        let msg = msg.to_str().ok().context("Expected JSON message")?;
```
matszczygiel, line 90: :+1:

drop-transfer/src/ws/server/v5.rs
```diff
@@ -87,7 +87,10 @@ impl<'a> handler::HandlerInit for HandlerInit<'a> {
             .await
             .context("Failed to receive transfer request")?;
 
-        let msg = msg.to_str().ok().context("Expected JOSN message")?;
+        // print msg as ascii
+        debug!(self.logger, "************** msg:\n\t{msg:?}");
```
matszczygiel, line 91:
```suggestion
        trace!(self.logger, "************** msg:\n\t{msg:?}");
```

drop-transfer/src/ws/server/mod.rs
LukasPukenis: thanks

drop-transfer/src/ws/server/v5.rs
LukasPukenis: I am planning on removing these in the final MR, just desperate to get it working at the moment

LukasPukenis: I am planning on removing these in the final MR, just desperate to get it working at the moment

drop-transfer/src/check.rs
LukasPukenis: My understanding of tokio::select! was lacking, you're totally correct on this one

drop-transfer/src/check.rs
```diff
@@ -7,55 +7,84 @@ use tokio_util::sync::CancellationToken;
...
+        debug!(
+            logger,
+            "Client job will restart in {:?} seconds for {}",
+            delay,
+            xfer.id()
+        );
     }
 
-    info!(logger, "Transfer {} is gone. Clearing", xfer.id());
+    if remove_xfer {
```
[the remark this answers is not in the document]
LukasPukenis: To address all the leftover comments:
- tokio::select! was really badly used and was causing job cancellations
- I've added `ControlFlow` to `run` methods so the retry loops would be aware if the task should be checked for retry or short-circuit.

It proved to be good measures as the testcases now pass:
```
Failure summary:
*** Scenario scenario27-7, failed containers: ['DROP_PEER_REN-scenario27-7']
*** Scenario scenario42-2, failed containers: ['DROP_PEER_REN-scenario42-2']
*** Test suite results: 103 scenarios, 2 failed. Succeeded (98.06%), on average one scenario took 3 seconds
```

drop-transfer/src/check.rs
```diff
@@ -7,56 +7,107 @@ use tokio_util::sync::CancellationToken;
...
+    refresh_trigger: tokio::sync::watch::Receiver<()>,
     state: Arc<State>,
     xfer: Arc<IncomingTransfer>,
     logger: Logger,
     guard: AliveGuard,
     stop: CancellationToken,
 ) -> tokio::task::JoinHandle<()> {
     let id = xfer.id();
-    let job = run(state, xfer, logger.clone());
+    let refresh_trigger = refresh_trigger;
```
matszczygiel, line 18:
```suggestion
```

drop-transfer/src/check.rs
```diff
@@ -7,56 +7,107 @@ use tokio_util::sync::CancellationToken;
...
+                            break;
+                        }
+                        ControlFlow::Continue(_) => {
+                            debug!(logger, "Check job continue: {id}");
+
+                            match refresh_trigger.has_changed() {
+                                Ok(true) => debug!(logger, "Check job refresh: {id}"),
+                                Ok(false) => {
+                                    debug!(logger, "Check job no refresh: {id}");
+                                    break;
```
matszczygiel, line 45: By breaking the loop here you're actually canceling the task, which should not happen here.

drop-transfer/src/check.rs
```diff
@@ -7,56 +7,107 @@ use tokio_util::sync::CancellationToken;
...
-        tokio::select! {
-            biased;
+            tokio::select! {
+                biased;
 
-            _ = stop.cancelled() => {
-                debug!(logger, "Check job stop for transfer: {id}");
-            },
-            _ = job => (),
+                _ = stop.cancelled() => {
```
matszczygiel, line 28: In general, you should have a different architecture here. There should be a task that runs the `run()` task on channel wakeup in a loop and here the task should be executed inside select with the stop token.
That said the select should have to entries (stop and the task) and shouldn't be in the loop.
The loop should be internal to the task and should not be mixed with the stopping token.

drop-transfer/src/manager.rs
```diff
@@ -998,6 +1004,46 @@ pub(crate) async fn restore_transfers_state(state: &Arc<State>, logger: &Logger)
     *state.transfer_manager.outgoing.lock().await = outgoing;
 }
 
+pub(crate) async fn resume(
+    refresh_trigger: tokio::sync::watch::Receiver<()>,
```
matszczygiel, line 1008: It's a minor but you don't need the full ownership here of the `Receiver`, meaning it can be passed by reference, as the stop token is for example

drop-transfer/src/service.rs
```diff
@@ -37,15 +36,13 @@ pub(super) struct State {
...
-type TaskRegistry = HashMap<uuid::Uuid, tokio::task::JoinHandle<()>>;
-
 pub struct Service {
     pub(super) state: Arc<State>,
     stop: CancellationToken,
     waiter: AliveWaiter,
     pub(super) logger: Logger,
 
-    tasks: TaskRegistry,
+    refresh_trigger: tokio::sync::watch::Sender<()>,
```
matszczygiel, line 45: Since the three things, namely `trigger`, `stop`, and `waiter` are almost always passed together, maybe it's worth wrapping them into a structure. Thus simplifying the arguments of many functions, like `ws::server::spawn`

matszczygiel: Oh, sorry the `Sender` cannot be cloned, so there can probably be a method producing the structure consisting of `stop`, `waiter`, and `trigger: Receiver`

LukasPukenis: That might be a worthwhile additional idea to revisit once this MR lands

drop-transfer/src/check.rs
LukasPukenis: I want to break the task if I have already finished it and there was no API signal to retry again and I want to spin the loop again in case I received API call while the task was running.

matszczygiel: Are you sure? I imagine you just want to wait for the channel trigger and do it again, not break the task

drop-transfer/src/check.rs
LukasPukenis: adjusted to changes

drop-transfer/src/check.rs
```diff
@@ -7,51 +7,91 @@ use tokio_util::sync::CancellationToken;
...
+                let cf = run(&state, xfer.clone(), &logger).await;
+
+                if cf.is_break() {
+                    break;
+                }
+
+                #[allow(clippy::let_underscore_must_use)]
+                let _ = refresh_trigger.changed().await;
+            }
+        };
```
matszczygiel, line 36:
```suggestion
        let task = async {
            loop {
                let cf = run(&state, xfer.clone(), &logger).await;

                if cf.is_break() {
                    break;
                }

                #[allow(clippy::let_underscore_must_use)]
                let _ = refresh_trigger.changed().await;
            }
        };
```

drop-transfer/src/check.rs
```diff
@@ -7,51 +7,91 @@ use tokio_util::sync::CancellationToken;
...
     guard: AliveGuard,
     stop: CancellationToken,
-) -> tokio::task::JoinHandle<()> {
+) {
     let id = xfer.id();
-    let job = run(state, xfer, logger.clone());
 
     tokio::spawn(async move {
         let _guard = guard;
+        let mut refresh_trigger = refresh_trigger;
```
matszczygiel, line 21:
```suggestion
```
and change the `spawn()` function parameter to be mutable `mut refresh_trigger: tokio::sync::watch::Receiver<()>,`

drop-transfer/src/check.rs
```diff
@@ -7,51 +7,91 @@ use tokio_util::sync::CancellationToken;
...
+        let task = async move {
+            let logger = logger_clone;
+            loop {
+                let cf = run(&state, xfer.clone(), &logger).await;
+
+                if cf.is_break() {
+                    break;
+                }
+
+                #[allow(clippy::let_underscore_must_use)]
```
matszczygiel, line 33: I think we've disabled this lint after our discussion under the rust 1.72.1 bump PR

drop-transfer/src/check.rs
```diff
@@ -7,51 +7,91 @@ use tokio_util::sync::CancellationToken;
...
+            _ = task => ()
         }
-    })
+    });
 }
 
-async fn run(state: Arc<State>, xfer: Arc<IncomingTransfer>, logger: Logger) {
-    loop {
-        tokio::time::sleep(drop_config::ALIVE_CHECK_INTERVAL).await;
+async fn run(state: &State, xfer: Arc<IncomingTransfer>, logger: &Logger) -> ControlFlow<()> {
```
matszczygiel, line 49:
```suggestion
async fn run(state: &State, xfer: &IncomingTransfer, logger: &Logger) -> ControlFlow<()> {
```

drop-transfer/src/ws/client/mod.rs
```diff
@@ -61,46 +62,89 @@ struct RunContext<'a> {
...
     let id = xfer.id();
-    let job = run(state, xfer, logger.clone(), guard.clone());
 
     tokio::spawn(async move {
-        let _guard = guard;
+        let mut refresh_trigger = refresh_trigger;
+        let logger_clone = logger.clone();
+
+        let task = async move {
+            let logger = logger_clone;
```
matszczygiel, line 85:
```suggestion
        let task = async {
```

drop-transfer/src/ws/client/mod.rs
```diff
@@ -61,46 +62,89 @@ struct RunContext<'a> {
...
-            warn!(
-                logger,
-                "Failed to clear sync state for {}: {err}",
-                xfer.id()
-            );
+async fn run(
+    state: Arc<State>,
+    xfer: Arc<OutgoingTransfer>,
+    logger: Logger,
+    alive: AliveGuard,
```
matszczygiel, line 112: You can use references here

drop-transfer/src/ws/client/mod.rs
```diff
@@ -179,7 +224,7 @@ async fn establish_ws_conn(
         Ok(sock) => sock,
         Err(err) => {
             debug!(logger, "Failed to connect: {:?}", err,);
-            return WsConnection::Recoverable;
+            return WsConnection::Stopped;
```
matszczygiel, line 227: Why do we introduce another error variant here? Isn't it just a `Recoverable` error?

drop-transfer/src/ws/server/mod.rs
```diff
@@ -351,6 +365,7 @@ async fn handle_rejection(
 struct RunContext<'a> {
     logger: &'a slog::Logger,
     state: Arc<State>,
+    refresh_trigger: &'a tokio::sync::watch::Receiver<()>,
```
matszczygiel, line 368: Do we use this field?

drop-transfer/src/check.rs
```diff
@@ -7,51 +7,91 @@ use tokio_util::sync::CancellationToken;
...
+            remove_xfer = true;
             break;
         }
+
+        debug!(
+            logger,
+            "Client job will restart in {:?} seconds for {}",
+            delay,
+            xfer.id()
+        );
```
Lipt0nas, line 75: Give that there's a debug log above, maybe this one is unneeded, or at the very least shouldn't proc on the last iteration since it won't actually restart, which could possibly cause confusion when someone reads the logs

drop-transfer/src/ws/client/mod.rs
LukasPukenis: It was actually present before. When debugging I restored it back and didn't look back at it

drop-transfer/src/check.rs
LukasPukenis: actually it was not necessary anymore, correct

LukasPukenis:
@Lipt0nas @matszczygiel github has re-request review button, so you're reviewing unfinished changes, just letting you know for the future

drop-transfer/src/ws/server/mod.rs
LukasPukenis: `init_manager` function uses it to spawn check tasks

Lipt0nas:
Id probably put it back into draft in this case, since we can't exactly know that it's in-progress otherwise

drop-transfer/src/check.rs
LukasPukenis: removed the `move`, it was unecessary

drop-transfer/src/check.rs
LukasPukenis: This one is not actually that easy, downstream calls require it to be Arc

drop-transfer/src/ws/client/mod.rs
LukasPukenis: actually i can

drop-transfer/src/ws/client/mod.rs
LukasPukenis: this was removed

drop-transfer/src/check.rs
LukasPukenis: that's nice, I was actually writing `a: mut T` and it wasn't compiling initially :D looks like the `mut` needs to appear on the left side

drop-transfer/src/check.rs
matszczygiel: So then
```suggestion
async fn run(state: &State, xfer: &Arc<IncomingTransfer>, logger: &Logger) -> ControlFlow<()> {
```
That should reduce the amount of cloning

drop-transfer/src/check.rs
```diff
@@ -7,51 +7,80 @@ use tokio_util::sync::CancellationToken;
...
         if !state.transfer_manager.is_incoming_alive(xfer.id()).await {
+            remove_xfer = true;
             break;
         }
 
-        if make_request(&state, &xfer, &logger).await.is_break() {
+        if make_request(state, &xfer, logger).await.is_break() {
+            remove_xfer = true;
             break;
         }
```
matszczygiel, line 64: Minor: you can do something like this
```rust
        if !state.transfer_manager.is_incoming_alive(xfer.id()).await {
            return ControlFlow::Break(());
            break;
        }

        make_request(state, &xfer, logger).await.is_break()?;
```
And move the transfer removal logic to the `let task = async {...}` in the `if cf.is_break() { here }```
It would be less line of code to comprehend but you can skip this, it's OK as it is.
Note: you can use `?` on `ControlFlow` to break early with `Break`, same as for `Err()`

drop-transfer/src/ws/client/mod.rs
```diff
@@ -66,41 +67,78 @@ enum WsConnection {
...
+            debug!(logger, "connection status is irrecoverable");
+            if let Err(err) = state.transfer_manager.outgoing_remove(xfer.id()).await {
+                warn!(
+                    logger,
+                    "Failed to clear sync state for {}: {err}",
+                    xfer.id()
+                );
+            }
+            return ControlFlow::Break(());
         }
```
matszczygiel, line 130: Minor: it could be the same where namely:
```rust
        connect_to_peer(&state, &xfer, logger, alive).await?;
```
And move the transfer removal block one function up the call stack

LukasPukenis:
@matszczygiel what do you think about the whole state of this PR? I fear that minor comments will not end :D

drop-transfer/src/ws/client/mod.rs
```diff
@@ -66,41 +67,78 @@ enum WsConnection {
...
+
+        debug!(
+            logger,
+            "connection status is recoverable for: {}, will sleep for {:?}",
+            xfer.id(),
+            delay
+        );
```
Lipt0nas, line 137: as in the previous situation, this log probably shouldn't be here anymore

LukasPukenis: yep

matszczygiel:
@LukasPukenis I will not push for the minor ones, it's basically ready.
But I'm waiting for your comment on them, thus not approving yet

drop-transfer/src/check.rs
LukasPukenis: I don't personally get used to passing Arcs as references, but this worked

drop-transfer/src/check.rs
LukasPukenis: I gave both comments a go and it produced a nicer code so I've commited it

drop-transfer/src/ws/client/mod.rs
LukasPukenis: I've changed both spawn sites to move the removal logic into the call site

matszczygiel (approved):

PR #248, opened 2023-10-09
build: Run cargo update

lcruz99:
Adding tvOS support requires some dependencies to be updated. In the process I've realized all the dependencies could be updated with a full `cargo update`, with no apparent issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,36 @@
 version = 3
 
 [[package]]
-name = "ahash"
-version = "0.7.6"
+name = "addr2line"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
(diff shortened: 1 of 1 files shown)


LukasPukenis:
Did you check this change against AppleTV support MR?

matszczygiel:
@lcruz99 we need to check if it builds with moose as we have had bad experiences with updating the deps, because the build can fail. Let's wait for the pipeline to finish

LukasPukenis:
I see that crono update is the same version as @Lipt0nas has in his MR so it should be good in that regard, no other crate has caused us issues other than chrono

Lipt0nas:
It failed on the current build of moose, once we update to the new one, the chrono version should match

matszczygiel:
Let's the merge moose update then

lcruz99:
> ou check this change against AppleTV support MR?

Locally, against the specific apple tvOS MR on CI I'm still working on it

LukasPukenis:
@lcruz99 leave this PR to us, we will merge @Lipt0nas moose changes, then we can merge this one as we will also do Apple stuff very soon :)

matszczygiel:
@Lipt0nas I've fixed the build, can we merge this one?

Lipt0nas (approved):
Yup, shouldn't be a problem

PR #250, opened 2023-10-09
Add timestamps to emmited events

Lipt0nas:


LukasPukenis (dismissed):
As discussed, good compromise to add ffi layer timestamps, looks clean

changelog.md
```diff
@@ -1,7 +1,8 @@
-### v6.0.0
-### **UNRELEASED**
+### v5.4.0
+### **Unforeseen Moose**
```
matszczygiel, line 2: Me like it :+1:

drop-transfer/src/utils.rs
```diff
@@ -149,6 +150,14 @@ pub fn to_kb(bytes: u64) -> i32 {
     (bytes as f64 / 1024.0).ceil() as i32
 }
 
+/// Get the current system timestamp in milliseconds, used for event timestamps
+pub fn current_timestamp() -> u64 {
+    SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("Failed to retrieve system time")
```
matszczygiel, line 157: Um, it probably would never fail but I'd go for `unwrap_or_default()` just in case

matszczygiel: BTW, why this is in the `drop-transfer` crate? It's used only in the `norddrop` crate

Lipt0nas: Made sense for me to put it here, since `norddrop` is an extension of `drop-transfer`, and it already has a `utils` file that conveniently houses utils, in addition to it being a good spot if we decided to move the timestamps to being created in internal events, as opposed to what we do now

Lipt0nas: Though i wouldn't be against moving it if you think it makes more sense to be in `norddrop` instead

matszczygiel: I mean things should be defined near the usage, though it's not too much of a deal.
However, it seems logical to keep it it here and use the timestamp on the internal events, or define it in the norddrop and use it only there. The mix of these two is a bit odd to me

matszczygiel: In any case, it's your call

matszczygiel:
@Lipt0nas now, as I think it makes sense to include the timestamp in the internal event. Because with the API events, you include the event after dequeuing them. I can imagine that the event sits for quite a long time in the channel thus making the timestamps biased

Lipt0nas:
Turned out to be a bit more of an intrusive change than expected, but isn't too bad

changelog.md
```diff
@@ -1,7 +1,13 @@
...
 ---
 * Update moose tracker to v2.0.0
+* Add `moose_app_version` field to config
+* Add timestamps to emitted events
+* Add DLL signing to Windows builds
```
matszczygiel, line 7: :bowing_man:

drop-transfer/examples/udrop.rs
```diff
@@ -148,17 +148,17 @@ fn print_event(ev: &Event) {
...
     storage: &Storage,
-    rx: &mut mpsc::UnboundedReceiver<Event>,
+    rx: &mut mpsc::UnboundedReceiver<(Event, u64)>,
     out_dir: &Path,
 ) -> anyhow::Result<()> {
     info!("Awaiting events…");
 
     let mut active_file_downloads = BTreeMap::new();
     let mut storage = drop_transfer::StorageDispatch::new(storage);
     while let Some(ev) = rx.recv().await {
```
matszczygiel, line 158:
```suggestion
    while let Some((ev, ts)) = rx.recv().await {
```

drop-transfer/examples/udrop.rs
```diff
@@ -382,15 +382,19 @@ async fn main() -> anyhow::Result<()> {
...
+    rx: &mut mpsc::UnboundedReceiver<(Event, u64)>,
+    storage: &Storage,
+) {
     info!("Stopping the service");
 
     service.stop().await;
     let mut storage = drop_transfer::StorageDispatch::new(storage);
 
     // Drain events
     while let Some(ev) = rx.recv().await {
```
matszczygiel, line 396: Same here

drop-transfer/src/service.rs
```diff
@@ -137,15 +138,21 @@ impl Service {
 
             self.state
                 .event_tx
-                .send(Event::OutgoingTransferFailed(xfer.clone(), err, true))
+                .send((
+                    Event::OutgoingTransferFailed(xfer.clone(), err, true),
+                    utils::current_timestamp(),
+                ))
                 .expect("Event channel should be open");
```
matszczygiel, line 145: TBH it is asking itself for a refactor into a separate method on `State::emit(&self, event: crate::Event)` and then the `utils::current_timestamp()` can be dropped altogether since it would be used in only one place if I'm not wrong

Lipt0nas: the `State::emit` does sound pretty nice, but the `utils` part would have to stay, since the Transfer Request/Queue events still need to use it

drop-transfer/src/ws/events.rs
```diff
@@ -73,7 +73,7 @@ impl<T: Transfer> FileEventTx<T> {
         }
 
         lock.tx
-            .send(event)
+            .send((event, utils::current_timestamp()))
```
matszczygiel, line 76: Oh, yeah. So `current_timestamp()` would still be used here.
This also begs for a `FileEventTxInner::emit(&self, event)` method

norddrop/src/device/mod.rs
```diff
@@ -120,15 +120,15 @@ impl NordDropFFI {
         let ed = self.event_dispatcher.clone();
         let event_logger = self.logger.clone();
         let event_storage = storage.clone();
-        let (tx, mut rx) = mpsc::unbounded_channel();
+        let (tx, mut rx) = mpsc::unbounded_channel::<(Event, u64)>();
```
matszczygiel, line 123: Very minor: I think `(Event, SystemTime)` would state the intent better. But you can skip this

drop-transfer/src/ws/events.rs
Lipt0nas: Question regarding both of the `emit` refactors - do we want to return an error from those functions or not? At first i thought that it would make sense to return an error, even if we `expect` all of them on the other side, at least we get a bit more info, on which event it failed to send, but the more i look at it, the more i think that this information is not really useful, and started leaning on returning void and just having a generic `expect` for the calls. What do you think?

matszczygiel: Yeah, go for void.
If the expect is triggered then it would indicate the severe logic flaw, with potentially far worse consequences than controllable panic in an `emit()` function.

norddrop/src/device/mod.rs
Lipt0nas: Can't say i really like where the acquisition of the timestamp would move with this change, will choose to skip this :)

LukasPukenis: just tagging along - `SystemTime` would definitely improve the semantics of this

Lipt0nas: It was wrong wording by me, i meant the calculation of the timestamp instead of the acquisition

i've changed it to `SystemTime` though

matszczygiel (approved):

PR #252, opened 2023-10-11
Server side authentication

matszczygiel:
Where it is, implemented according to the RFC.
I have a question though. Is it OK for it to go into the `dev` branch?


drop-transfer/src/auth.rs
```diff
@@ -33,10 +33,33 @@ impl Context {
...
+        let ticket = response
+            .headers()
+            .get(drop_auth::http::Authorization::KEY)
+            .context("Missing 'authorization' header")?
+            .to_str()
+            .context("Invalid 'authorization' header value")?;
+
+        anyhow::ensure!(
+            self.authorize(ip, ticket, nonce),
+            "Cannot authrozie server with ticket provided"
```
Lipt0nas, line 53:
```suggestion
            "Cannot authorize server with ticket provided"
```

matszczygiel: done

Lipt0nas (approved):

PR #254, opened 2023-10-11
fix clippy 1.73 lints

matszczygiel:


drop-transfer/src/ws/server/v2.rs
```diff
@@ -426,11 +427,14 @@ impl handler::Downloader for Downloader {
...
+
         let suffix: String = suffix
             .finalize()
             .iter()
-            .map(|b| format!("{:02x}", b))
-            .collect();
+            .fold(String::new(), |mut output, b| {
+                let _ = write!(output, "{b:02x}");
+                output
+            });
```
LukasPukenis, line 437: What was the error?

Lipt0nas: it was that saying that what was written before could be done more efficiently

i guess this could raise a question if we should be developing with the same version of rust as we're building with

matszczygiel: The warning was:
```
warning: use of `format!` to build up a string from an iterator
   --> drop-transfer/src/ws/server/v2.rs:429:30
    |
429 |           let suffix: String = suffix
    |  ______________________________^
430 | |             .finalize()
431 | |             .iter()
432 | |             .map(|b| format!("{:02x}", b))
433 | |             .collect();
    | |______________________^
    |
help: call `fold` instead
   --> drop-transfer/src/ws/server/v2.rs:432:14
    |
432 |             .map(|b| format!("{:02x}", b))
    |              ^^^
help: ... and use the `write!` macro here
   --> drop-transfer/src/ws/server/v2.rs:432:22
    |
432 |             .map(|b| format!("{:02x}", b))
    |                      ^^^^^^^^^^^^^^^^^^^^
    = note: this can be written more efficiently by appending to a `String` directly
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#format_collect
    = note: `#[warn(clippy::format_collect)]` on by default
    ```

Comment (matszczygiel) [reply drop-transfer/src/ws/server/v2.rs:437]: The warning was:
```
warning: use of `format!` to build up a string from an iterator
   --> drop-transfer/src/ws/server/v2.rs:429:30
    |
429 |           let suffix: String = suffix
    |  ______________________________^
430 | |             .finalize()
431 | |             .iter()
432 | |             .map(|b| format!("{:02x}", b))
433 | |             .collect();
    | |______________________^
    |
help: call `fold` instead
   --> drop-transfer/src/ws/server/v2.rs:432:14
    |
432 |             .map(|b| format!("{:02x}", b))
    |              ^^^
help: ... and use the `write!` macro here
   --> drop-transfer/src/ws/server/v2.rs:432:22
    |
432 |             .map(|b| format!("{:02x}", b))
    |                      ^^^^^^^^^^^^^^^^^^^^
    = note: this can be written more efficiently by appending to a `String` directly
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#format_collect
    = note: `#[warn(clippy::format_collect)]` on by default
    ```

matszczygiel: @Lipt0nas Why we shouldn't? Newer clippy versions have more, better lints. What's wrong with using those?

matszczygiel: Going to https://rust-lang.github.io/rust-clippy/master/index.html#/format_collect it's hard to disagree with the lint, so I applied it

LukasPukenis: @Lipt0nas developing and building with the same version makes most sense, it reduces friction :)

matszczygiel: I kind of disagree. The reason for a specific version in our case is the LLVM version requirement or whatever. I still don't get why we cannot benefit from the newest lints though.
These are supposed to make the product better

Lipt0nas: We do have CI that actually builds the final binary with the version we should use, so i guess it doesn't really matter what version we use internally, worst case we'd just have to change it back if some unstable API is used

matszczygiel: I mean I'm talking specifically about the clippy, not about the rustc. It just happens they are versioned with the same number

Lipt0nas: oh, so clippy isn't tied to a particular Rust version?

matszczygiel: They're released together, hence the version. But these are different projects, same with rust-analyzer for example.

Lipt0nas (approved):
Since clippy isn't tied to a Rust version, this is not a problem to merge

PR #255, opened 2023-10-12
Ensure the event task finishes before norddrop_stop() returns

matszczygiel:


norddrop/src/device/mod.rs
```diff
@@ -178,7 +189,11 @@ impl NordDropFFI {
             .take()
             .ok_or(ffi::types::NORDDROP_RES_NOT_STARTED)?;
 
-        self.rt.block_on(instance.stop());
+        self.rt.block_on(async {
+            instance.service.stop().await;
+            let _ = instance.event_task.await;
+        });
+
```
LukasPukenis, line 196: It looks like it should deadlock as `event_task` depends on `rx` which depends on `service.tx` which is dropped only when `instance` is dropped, But yet it seems to work, could you explain why?

matszczygiel: Huh, good point. But why doesn't it deadlock then

matszczygiel: Oh, yeah. It doesn't deadlock because `service.stop()` takes service by value. So, there are no problems here.
Somebody was wise enough to take `self` by value in `Service::stop()`

matszczygiel:
@LukasPukenis Can you run this branch on your machine? I see some tests failing but with very clear errors, rather not related to the problem you mentioned about scenario44-4

LukasPukenis (approved):
it's nice

PR #258, opened 2023-10-13
Add scenario6-2(additional dir test)

LukasPukenis:
Testing that sending directory and
downloading it into different destinations
will produce the trees in those destinations
as well.


LukasPukenis:
basically it's a derivative scenario6 as I was curious what would happen when writing documentation

matszczygiel (approved):

PR #259, opened 2023-10-13
Remove pending mentions from outgoing transfer

LukasPukenis:
The example in API call was containing
`outgoing` state for the transfer which we don't have anymore


norddrop/ffi/bindings/norddrop.h
```diff
@@ -405,7 +401,7 @@ enum norddrop_result norddrop_purge_transfers_until(const struct norddrop *dev,
  *             "by_peer": false
  *         }
  *     ],
- *     "type": "outgoing",
```
LukasPukenis, line 408: This was actually a bug, in the example, most probably the same entry from Outgoing was copied

matszczygiel (approved):

PR #263, opened 2023-10-13
Fix: There is a lack of information in the UI about illegal file names being normalized

matszczygiel:
Implemented as agreed on. The drawback is that the transfer is preserved across upgrades from the older protocol to the new one, but I've noticed the problem is there even without normalization so I guess we don't break additional stuff with this PR.

Probably best to review this PR commit by commit


Lipt0nas:
Looks good overall, the only thing that still needs be to be done is to update the `19-2` scenario's description and expected events

matszczygiel:
@Lipt0nas I've pushed a change for scenario19-2

Lipt0nas (approved):

PR #265, opened 2023-10-17
fix flaky tests

matszczygiel:
This PR fixes a flakiness in some tests, plus some logic errors:
* Improve the reconnection logic, so that the trigger is not ignored while in the backoff loop
* There was a bug where the file was stopped for good on disconnection

I've also split the analytics tests from the regular ones, for easier inspection

### DIFF
--- a/drop-transfer/src/ws/client/mod.rs
+++ b/drop-transfer/src/ws/client/mod.rs
@@ -10,7 +10,6 @@ use std::{
     net::{IpAddr, SocketAddr},
     ops::ControlFlow,
     sync::Arc,
-    time::Duration,
 };
 
 use anyhow::Context;
@@ -73,7 +72,7 @@ enum RequestError {
 }
 
 pub(crate) fn spawn(
-    mut refresh_trigger: tokio::sync::watch::Receiver<()>,
+    refresh_trigger: tokio::sync::watch::Receiver<()>,
     state: Arc<State>,
     xfer: Arc<OutgoingTransfer>,
     logger: Logger,
@@ -83,9 +82,11 @@ pub(crate) fn spawn(
     let id = xfer.id();
 
     tokio::spawn(async move {
+        let mut backoff = utils::RetryTrigger::new(refresh_trigger);
+
         let task = async {
             loop {
-                let cf = run(state.clone(), xfer.clone(), &logger, &guard).await;
+                let cf = connect_to_peer(&state, &xfer, &logger, &guard).await;
                 if cf.is_break() {
                     debug!(logger, "connection status is irrecoverable");
                     if let Err(err) = state.transfer_manager.outgoing_remove(xfer.id()).await {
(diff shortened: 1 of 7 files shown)


drop-transfer/src/ws/client/mod.rs
```diff
@@ -114,34 +115,14 @@ pub(crate) fn spawn(
...
-    ControlFlow::Continue(())
-}
-
 async fn connect_to_peer(
     state: &Arc<State>,
     xfer: &Arc<OutgoingTransfer>,
     logger: &Logger,
     alive: &AliveGuard,
 ) -> ControlFlow<()> {
+    debug!(logger, "Outcoming transfer job started for {}", xfer.id(),);
```
Lipt0nas, line 124:
```suggestion
    debug!(logger, "Outgoing transfer job started for {}", xfer.id(),);
```

test/scenarios.py
```diff
@@ -10145,7 +10145,7 @@
         },
     ),
     Scenario(
-        "scenario-analytics-1",
+        "analytics-1",
```
Lipt0nas, line 10148: I really like the idea of splitting analytics tests into a separate job

There is still a question for scenarios `26-1` and `26-2`, they are also testing analytics events, though they also have some other functional testing alongside it, i would lean towards moving them to the `analytics` space as well, what is your opinion?

drop-transfer/src/ws/server/mod.rs
```diff
@@ -905,15 +905,7 @@ impl FileXferTask {
...
                         Err(crate::Error::Canceled) => {
-                            info!(logger, "File {} cancelled", self.file.id());
-
-                            if let Err(err) = state
-                                .transfer_manager
-                                .incoming_download_cancel(self.xfer.id(), self.file.id())
-                                .await
-                            {
-                                warn!(logger, "Failed to store download finish: {err}");
-                            }
```
Lipt0nas, line 916: Could you explain on how this was a bug?

Lipt0nas (commented):
Really clean refactor of the wait loop :ok_hand:

test/scenarios.py
matszczygiel: Yeah, sure. Good point :+1:

drop-transfer/src/ws/client/mod.rs
matszczygiel: done

test/scenarios.py
matszczygiel: Moved

drop-transfer/src/ws/server/mod.rs
Lipt0nas: Ah i see, thank you for explaining

True about QA, perhaps they're still doing more surface tests

Lipt0nas (approved):
Damn is it nice to see dev pipelines being fully green (even if still slightly flaky)

PR #266, opened 2023-10-18
tests: reduce flakyness of the test cases

matszczygiel:
I've tried to run the workflow several times by now and it's more or less green

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -907,7 +907,7 @@ scenarios = [
                             },
                         )
                     ),
-                    action.SleepMs(200),
+                    action.Sleep(0.2),
                     action.CancelTransferRequest([0]),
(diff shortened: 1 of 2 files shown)


test/scenarios.py
```diff
@@ -5419,7 +5419,7 @@
                             0, FILES["testfile-small"].id, Error.FILE_REJECTED
                         ),
                     ),
-                    action.SleepMs(200),
```
LukasPukenis, line 5422: minor: I think `SleepMs()` is a bit redundant, `Sleep()` is enough to be had in the codebase, especially when we can pass `0.2` instead of `200`

LukasPukenis (approved):

LukasPukenis (approved):

PR #270, opened 2023-10-19
Remove TODO comments

matszczygiel:
I've removed todo which seemed unrealistic or redundant to me


LukasPukenis (approved):
That's a good cleanup. Overall I do enjoy TODO comments but when unaddressed they become stale

PR #271, opened 2023-10-19
Recude flakyness of analytics tests

Lipt0nas:
In multi-file transfers, it's not worth to pendatically check for specific file ID's we already know that the batch event contains the correct information, so it should be safe enough, this was the main cause of flakyness, in addition, the check for protocol version has also been made pedantic, as this could in theory help catch a super obscure bug and isn't too big of pain to update, given the frequency of new protocol versions coming out

And also moved all of the analytics tests to be close together

### DIFF
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -4818,210 +4818,6 @@ scenarios = [
         },
     ),
     Scenario(
-        "analytics-6-1",
-        "Test if the instance can recover on database corruption",
-        {
-            "DROP_PEER_REN": ActionList(
-                [
-                    action.Start(
-                        "DROP_PEER_REN",
-                        dbpath="/tmp/db/26-1-corrupted.sqlite",
-                    ),
-                    action.Wait(event.RuntimeError(Error.DB_LOST)),
-                    action.NoEvent(),
-                    action.Stop(),
-                    action.AssertMooseEvents(
-                        [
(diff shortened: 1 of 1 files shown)


test/scenarios.py
```diff
@@ -10325,6 +10325,11 @@
                             "app_version": "test-framework",
                             "prod": false,
                             "init_duration": ">0"
+                        }""",
+                            """{
+                            "type": "transfer_state",
+                            "protocol_version": "6",
+                            "result": 0
```
LukasPukenis, line 10332: It appears to be just moved around but is no functional chance?

LukasPukenis (dismissed):

test/scenarios.py
Lipt0nas: the order might have gotten messed up after some rebasing, but moved it here to follow the same layout - events expected in the order that they should arrive in

also, this is still a draft :D

LukasPukenis: aw shiiiiii :D

Lipt0nas:
After running the pipelines here on GitHub for over 20 times, none of them failed, should be safe to say that the tests have been unflaked at least somewhat

matszczygiel (approved):
@Lipt0nas Don't worry, it will fail the 21st time :)

PR #276, opened 2023-11-20
Allow for test running via label

LukasPukenis:
Previous label of `run tests` was obsolete as it actually didn't trigger tests in gitlab but rather internal stuff, so I renamed it.

I also added the same workflow to unit and integration tests. It should trigger the tests after adding the label of `run tests`

### DIFF
--- a/.github/workflows/gitlab.yml
+++ b/.github/workflows/gitlab.yml
@@ -11,7 +11,7 @@ permissions: {}
 jobs:
   trigger-gitlab-pipeline:
     runs-on: [self-hosted, gitlab]
-    if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'run tests')) }}
+    if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'gitlab')) }}
     steps:
       - uses: NordSecurity/trigger-gitlab-pipeline@v1
         with:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -2,6 +2,7 @@ name: Integration tests
 run-name: ${{ github.actor }} is running integration tests
 on:
   - push
+  - pull_request_target
 jobs:
   tests:
     runs-on: ubuntu-latest    
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -2,6 +2,7 @@ name: Unit tests
 run-name: ${{ github.actor }} is running unit tests
 on:
   - push
+  - pull_request_target
 jobs:
   tests:
     runs-on: ubuntu-latest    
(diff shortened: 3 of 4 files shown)


matszczygiel (dismissed):
Noice :+1:

.github/workflows/gitlab.yml
```diff
@@ -11,7 +11,7 @@ permissions: {}
 jobs:
   trigger-gitlab-pipeline:
     runs-on: [self-hosted, gitlab]
-    if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'run tests')) }}
+    if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'gitlab')) }}
```
Lipt0nas, line 14: True that this doesn't exactly reflect what is happening under the hood, but would there really be a scenario where you would want to trigger tests but not gitlab, or vice versa? i would maybe lean towards both to be triggered by one label

matszczygiel: I'd say the test should be run in all cases like it was (almost) before. The gitlab CI runs only when we (admins) attach a label to the PR.

Lipt0nas: do you mean tests should be ran for everyone, regardless of the label?

matszczygiel: Yes. Do you have an argument against it?

Lipt0nas: Since the tests are not running on our internal runners, i guess yeah, don't see why we can't run these automatically

Lipt0nas: Specially given that the analytics tests portion didn't receive the same treatment here :D

.github/workflows/unit-tests.yaml
```diff
@@ -2,9 +2,11 @@ name: Unit tests
 run-name: ${{ github.actor }} is running unit tests
 on:
   - push
+  - pull_request_target
```
matszczygiel, line 5: This one should go to the analytics as well

.github/workflows/unit-tests.yaml
```diff
@@ -2,9 +2,11 @@ name: Unit tests
...
 jobs:
   tests:
     runs-on: ubuntu-latest    
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'run tests')) }}
```
matszczygiel, line 9: Remove

.github/workflows/integration-tests.yml
```diff
@@ -2,9 +2,11 @@ name: Integration tests
 run-name: ${{ github.actor }} is running integration tests
 on:
   - push
+  - pull_request_target
 jobs:
   tests:
     runs-on: ubuntu-latest    
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'run tests')) }}
```
matszczygiel, line 9: Remove

PR #277, opened 2023-11-21
Implement Checksum events after file download

LukasPukenis:
New Events:
* ChecksumStarted
* ChecksumProgress
* ChecksumFinished

These events are emited when the file download finished just before the final TrasnferFinished event.

New config options:
* checksum_events_size_threshold_byttes - only emit if the file is equal or bigger than this amount.

### DIFF
--- a/drop-config/src/lib.rs
+++ b/drop-config/src/lib.rs
@@ -11,6 +11,8 @@ pub struct DropConfig {
     pub dir_depth_limit: usize,
     pub transfer_file_limit: usize,
     pub storage_path: String,
+    // If set the checksum events will be emited for every file of this or bigger size
+    pub checksum_events_size_threshold: Option<usize>,
 }
 
 impl Default for DropConfig {
@@ -19,6 +21,7 @@ impl Default for DropConfig {
             dir_depth_limit: 5,
             transfer_file_limit: 1000,
             storage_path: "libdrop.sqlite".to_string(),
+            checksum_events_size_threshold: None,
         }
     }
 }
--- a/drop-transfer/src/file/mod.rs
+++ b/drop-transfer/src/file/mod.rs
@@ -4,6 +4,7 @@ mod reader;
 
 use std::{
     fmt,
+    future::Future,
     io::{self, Read, Write},
     path::{Path, PathBuf},
 };
@@ -296,25 +297,33 @@ impl FileToSend {
     }
 
     /// Calculate sha2 of a file. This is a blocking operation
-    pub(crate) async fn checksum(
+    pub(crate) async fn checksum<F, Fut>(
         &self,
         limit: u64,
-        progress_tx: Option<tokio::sync::watch::Sender<u64>>,
-    ) -> crate::Result<[u8; 32]> {
+        progress_cb: Option<F>,
+    ) -> crate::Result<[u8; 32]>
+    where
+        F: FnMut(u64) -> Fut + Send + Sync,
+        Fut: Future<Output = ()>,
+    {
         let reader = reader::open(&self.source)?;
         let mut reader = io::BufReader::new(reader).take(limit);
-        let csum = checksum(&mut reader, progress_tx).await?;
+        let csum = checksum(&mut reader, progress_cb).await?;
         Ok(csum)
     }
 }
 
-pub async fn checksum(
+pub async fn checksum<F, Fut>(
     reader: &mut impl io::Read,
-    progress_tx: Option<tokio::sync::watch::Sender<u64>>,
-) -> io::Result<[u8; 32]> {
+    mut progress_cb: Option<F>,
+) -> io::Result<[u8; 32]>
+where
+    F: FnMut(u64) -> Fut + Send + Sync,
+    Fut: Future<Output = ()>,
+{
     let mut csum = sha2::Sha256::new();
 
-    let mut buf = vec![0u8; 8 * 1024]; // 8kB buffer
+    let mut buf = vec![0u8; 256 * 1024];
 
     let mut total_n: u64 = 0;
     loop {
@@ -326,10 +335,9 @@ pub async fn checksum(
         csum.write_all(&buf[..n])?;
 
         total_n += n as u64;
-        if let Some(ref tx) = progress_tx {
-            let _ = tx.send(total_n);
+        if let Some(progress_cb) = progress_cb.as_mut() {
+            progress_cb(total_n).await;
         }
-
         // Since these are all blocking operation we need to give tokio runtime a
         // timeslice
         tokio::task::yield_now().await;
@@ -362,7 +370,12 @@ mod tests {
 
     #[tokio::test]
     async fn checksum() {
-        let csum = super::checksum(&mut &TEST[..], None).await.unwrap();
+        let csum = super::checksum(
+            &mut &TEST[..],
+            None::<fn(u64) -> futures::future::Ready<()>>,
+        )
+        .await
+        .unwrap();
         assert_eq!(csum.as_slice(), EXPECTED);
     }
 
@@ -376,7 +389,9 @@ mod tests {
 
             let size = TEST.len() as _;
             let file = super::FileToSend::from_path(tmp.path(), size).unwrap();
-            file.checksum(size, None).await.unwrap()
+            file.checksum(size, None::<fn(u64) -> futures::future::Ready<()>>)
+                .await
+                .unwrap()
         };
 
         assert_eq!(csum.as_slice(), EXPECTED);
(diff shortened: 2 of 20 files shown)


drop-config/src/lib.rs
```diff
@@ -11,6 +11,7 @@ pub struct DropConfig {
     pub dir_depth_limit: usize,
     pub transfer_file_limit: usize,
     pub storage_path: String,
+    pub checksum_events_size_threshold: Option<usize>,
```
matszczygiel, line 14: Can you document what `None` means in this case?

drop-transfer/src/file/mod.rs
```diff
@@ -296,19 +296,27 @@ impl FileToSend {
...
-        let csum = checksum(&mut reader).await?;
+        let csum = checksum(&mut reader, progress_tx).await?;
         Ok(csum)
     }
 }
 
-pub async fn checksum(reader: &mut impl io::Read) -> io::Result<[u8; 32]> {
+pub async fn checksum(
+    reader: &mut impl io::Read,
+    progress_tx: Option<tokio::sync::watch::Sender<u64>>,
```
matszczygiel, line 313: IMHO it should be more generic and take a progress callback, e.g.
```rust
fn checksum(reader: &mut ..., progress_cb: impl Fn(u64))
```

drop-transfer/src/ws/server/mod.rs
```diff
@@ -763,7 +765,36 @@ impl FileXferTask {
                 return Err(crate::Error::UnexpectedData);
             }
 
-            downloader.validate(tmp_loc).await?;
+            if emit_checksum_events {
+                events.checksum_start().await;
+
+                let progress_chan = tokio::sync::watch::channel(0u64).0;
+
+                tokio::spawn({
```
matszczygiel, line 773: Why separate tasks for events? The flow is like this:
`-> oneshot channel send -> oneshot channel recv -> event channel send ->`
I'd say it can be shortened to
`-> event channel sed ->`

test/runner.py
```diff
@@ -17,10 +17,6 @@
 TESTCASE_TIMEOUT = 100
 SCENARIOS_AT_ONCE = 100
 
-STDERR_ERR_PATTERNS = [
```
matszczygiel, line 20: Why is it deleted?

test/scenarios.py
```diff
@@ -11018,4 +11018,78 @@
...
+                        "/tmp/received",
+                    ),
+                    action.Wait(event.Start(0, FILES["testfile-big"].id)),
+                    action.Wait(event.ChecksumStarted(0, FILES["testfile-big"].id)),
+                    action.Wait(event.ChecksumFinished(0, FILES["testfile-big"].id)),
+                    action.Wait(
+                        event.FinishFileDownloaded(
+                            0,
+                            FILES["testfile-big"].id,
+                            "/tmp/received/testfile-big",
```
matszczygiel, line 11078:
```suggestion
                            "/tmp/received/49/testfile-big",
```
The files from different tests can interfere with each other

matszczygiel:
I see one problem though. The sender (WS client) still does not get the checksumming progress info and is stuck at 100%. However, this cannot be fixed without the protocol changes

test/runner.py
LukasPukenis: This wasn't used anymore. Instead of fixing it in this PR I opted out for deleting it

drop-config/src/lib.rs
LukasPukenis: Sure, I will add a comment. Responding to the question - apps asked for a configurable option for the threshold, so not to introduce a breaking change, there's an Option as it might not even be there.

drop-transfer/src/file/mod.rs
LukasPukenis: I actually went the callback route initially but that deemed to be very difficult(to me) as the callback needs to issue async operations, thus I opted out for a channel instead.

test/scenarios.py
LukasPukenis: Good point, thanks for pointing out

LukasPukenis:
@matszczygiel regarding the sender - you're totally right. I thought about this myself and there's nothing we can do without augmenting the protocol.

drop-transfer/src/ws/server/mod.rs
LukasPukenis: I wanted to add some throttling in between so it wouldn't be emitted for every 8K increment, however in hindsight I could have just increased the checksumming buffer.

I am fine with both ways personally, this one I prefer because it also allows for isolated channel just for the progress instead of allowing sending any event by passing the regular channel

drop-transfer/src/file/mod.rs
LukasPukenis: Additional commit contains the conversion to a callback. Actually I've had success, even though it's was much more annoying to implement

drop-transfer/src/ws/server/mod.rs
LukasPukenis: Last commit contains the rework, now it's callback based and the throttling is done by increasing checksum size to 256kb

drop-transfer/src/file/mod.rs
```diff
@@ -12,6 +12,7 @@ use std::{os::unix::prelude::*, sync::Arc};
 
 use drop_analytics::TransferDirection;
 use drop_config::DropConfig;
+use futures_util::Future;
```
matszczygiel, line 15: `use std::future::Future;`

drop-transfer/src/file/mod.rs
```diff
@@ -296,19 +297,35 @@ impl FileToSend {
     }
 
     /// Calculate sha2 of a file. This is a blocking operation
-    pub(crate) async fn checksum(&self, limit: u64) -> crate::Result<[u8; 32]> {
+    pub(crate) async fn checksum<F, Fut>(
+        &self,
+        limit: u64,
+        progress_cb: Option<F>,
```
matszczygiel, line 303: Why option? You could pass an empty function if you want to ignore the progress report

drop-transfer/src/file/mod.rs
LukasPukenis: What is the difference?

drop-transfer/src/ws/server/handler.rs
```diff
@@ -1,5 +1,6 @@
 use std::{fs, path::PathBuf, sync::Arc, time::Duration};
 
+use futures_util::Future;
```
matszczygiel, line 3: `std::future::Future;`

drop-transfer/src/ws/server/v2.rs
```diff
@@ -11,6 +11,7 @@ use std::{
 use anyhow::Context;
 use drop_config::DropConfig;
 use drop_core::Status;
+use futures_util::Future;
```
matszczygiel, line 14:
```suggestion
use std::future::Future;
```

drop-transfer/src/ws/server/v4.rs
```diff
@@ -4,6 +4,7 @@ use anyhow::Context;
 use async_cell::sync::AsyncCell;
 use drop_config::DropConfig;
 use drop_core::Status;
+use futures_util::Future;
```
matszczygiel, line 7: same

drop-transfer/src/ws/server/v6.rs
```diff
@@ -4,6 +4,7 @@ use anyhow::Context;
 use async_cell::sync::AsyncCell;
 use drop_config::DropConfig;
 use drop_core::Status;
+use futures_util::Future;
```
matszczygiel, line 7:
```suggestion
use std::future::Future;
```

norddrop/src/device/types.rs
```diff
@@ -123,6 +141,9 @@ pub struct Config {
     pub moose_event_path: String,
     pub moose_prod: bool,
     pub storage_path: String,
+
+    #[serde(rename = "checksum_events_size_threshold_bytes")]
```
matszczygiel, line 145: Can we actually name the field like this?

LukasPukenis: In the codebase we for sure know we're ever talking in bytes, however I noticed in the API usage it sometimes raises the questions, thus I opted out for the best from both worlds.

It's however not a hard opinion, we can for sure have the same name in the codebase as well if you prefer

LukasPukenis:
@matszczygiel i've removed Arc from Events

norddrop/src/device/types.rs
matszczygiel: Nah, it's fine. I was just curious why it's like that

matszczygiel (approved):

PR #279, opened 2023-11-23
Lukasp/file 541 checksum events before download begins

LukasPukenis:
Introduce downloader side checksumming events when the transfer is being resumed.

For this the previous checksum event was reused but size was added so API user could calculate the progress properly, as well as additional state was added to the file so events could be emited properly.

### DIFF
--- a/drop-transfer/src/ws/events.rs
+++ b/drop-transfer/src/ws/events.rs
@@ -22,6 +22,7 @@ struct FileEventTxInner {
 enum State {
     Idle,
     Throttled,
+    Preflight,
     InFlight { started: Instant },
 }
 
@@ -74,7 +75,9 @@ impl<T: Transfer> FileEventTx<T> {
     async fn emit_in_flight(&self, event: Event) {
         let mut lock = self.inner.lock().await;
 
-        if !matches!(lock.state, State::InFlight { .. }) {
+        if !(matches!(lock.state, State::Preflight { .. })
+            || matches!(lock.state, State::InFlight { .. }))
+        {
             return;
         }
 
@@ -107,6 +110,7 @@ impl<T: Transfer> FileEventTx<T> {
             State::Idle => return,
             State::Throttled => Duration::ZERO,
             State::InFlight { started } => started.elapsed(),
+            State::Preflight => Duration::ZERO,
         };
 
         let phase = match event {
@@ -143,6 +147,7 @@ impl<T: Transfer> FileEventTx<T> {
             State::Idle => Duration::ZERO,
             State::Throttled => Duration::ZERO,
             State::InFlight { started } => started.elapsed(),
+            State::Preflight => Duration::ZERO,
         };
 
         let phase = match event {
@@ -179,6 +184,7 @@ impl<T: Transfer> FileEventTx<T> {
             State::Idle => None,
             State::Throttled => Some(Duration::ZERO),
             State::InFlight { started } => Some(started.elapsed()),
+            State::Preflight => Some(Duration::ZERO),
         };
 
         if let Some(elapsed) = elapsed {
@@ -202,10 +208,11 @@ impl<T: Transfer> FileEventTx<T> {
 }
 
 impl FileEventTx<IncomingTransfer> {
-    pub async fn checksum_start(&self) {
+    pub async fn checksum_start(&self, size: u64) {
         self.emit_in_flight(crate::Event::ChecksumStarted {
             transfer_id: self.xfer.id(),
             file_id: self.file_id.clone(),
+            size,
         })
         .await
     }
@@ -244,6 +251,11 @@ impl FileEventTx<IncomingTransfer> {
             offset,
         )])
         .await
+    }
+
+    pub async fn preflight(&self) {
+        let mut lock = self.inner.lock().await;
+        lock.state = State::Preflight;
     }
 
     pub async fn failed(&self, err: crate::Error) {
@@ -337,6 +349,7 @@ impl FileEventTx<OutgoingTransfer> {
             }
             State::Throttled => (),
             State::InFlight { .. } => (),
+            State::Preflight => (),
         }
     }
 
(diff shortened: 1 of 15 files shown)


drop-transfer/src/ws/events.rs
```diff
@@ -246,6 +258,10 @@ impl FileEventTx<IncomingTransfer> {
         .await
     }
 
+    pub async fn preflight(&self) {
+        self.preflight_inner().await
```
matszczygiel, line 262: The `fn preflight_inner()` is used only once here. Can you inline it?

drop-transfer/src/ws/server/v6.rs
```diff
@@ -580,8 +585,38 @@ impl handler::Downloader for Downloader {
...
+        } else {
+            None
+        };
+
+        let res = super::TmpFileState::load(&tmp_location.0, cb).await;
+
+        if will_emit_checksum_events {
+            events.checksum_finish().await;
+        }
+
```
matszczygiel, line 617: TBH, I somewhat have a problem with this logic here. It's shared among two versions and clutters the `Download` trait significantly.
I suggest expanding the `DownloadInit` enum into two cases: resuming and starting from the beginning and offloading the logic higher in the call stack

drop-transfer/src/ws/server/v4.rs
```diff
@@ -527,8 +532,34 @@ impl handler::Downloader for Downloader {
...
+            events.checksum_start(size).await;
+
+            Some({
+                let ev = events.clone();
+                move |progress: u64| {
+                    let ev = ev.clone();
+                    async move {
+                        ev.checksum_progress(progress).await;
+                    }
+                }
```
matszczygiel, line 553: These clones are not needed IMHO

drop-transfer/src/ws/server/v4.rs
```diff
@@ -515,7 +515,12 @@ impl Downloader {
 
 #[async_trait::async_trait]
 impl handler::Downloader for Downloader {
-    async fn init(&mut self, task: &super::FileXferTask) -> crate::Result<handler::DownloadInit> {
+    async fn init(
+        &mut self,
+        task: &super::FileXferTask,
+        events: Arc<FileEventTx<IncomingTransfer>>,
```
matszczygiel, line 521: I'm 99% sure this can be done by passing `events` by reference

drop-transfer/src/ws/events.rs
LukasPukenis: This requires making `preflight_inner()` or equivalent to be public which I don't really like

drop-transfer/src/ws/server/v4.rs
LukasPukenis: It emits
```
`events` does not live long enough
borrowed value does not live long enough
```

LukasPukenis: in the `start_download()` in the end if I pass in a reference here. This is due to the fact that the events is passed down from elsewhere and it comes in as Arc.

drop-transfer/src/ws/server/v4.rs
LukasPukenis: There must be the clones or else the closure is `FnOnce` as it moves the existing `ev` into it's scope

drop-transfer/src/ws/server/v6.rs
LukasPukenis: I've gained some simplification by reducing changes to v4 and v6 files by moving the logic one layer above

drop-transfer/src/ws/events.rs
matszczygiel: Um, I don't get what you mean. My point was: that since `_inner` is used only once (here), then we don't need it as a separate function. Is it used somewhere else as well?

LukasPukenis: Oh I get it now. Misunderstood at first. Inlined it

drop-transfer/src/ws/server/v4.rs
```diff
@@ -515,26 +516,15 @@ impl Downloader {
...
-                );
+    async fn init(
+        &mut self,
+        task: &super::FileXferTask,
+        tmpstate: Option<TmpFileState>,
+    ) -> crate::Result<handler::DownloadInit> {
+        match tmpstate {
+            Some(t) => {
+                let meta = &t.meta;
+                let csum = t.csum;
```
matszczygiel, line 527:
```suggestion
            Some(TmpFileState{ meta, csum} ) => {
```

drop-transfer/src/ws/server/v6.rs
```diff
@@ -568,26 +569,15 @@ impl Downloader {
...
-                );
+    async fn init(
+        &mut self,
+        task: &super::FileXferTask,
+        tmpstate: Option<TmpFileState>,
+    ) -> crate::Result<handler::DownloadInit> {
+        match tmpstate {
+            Some(t) => {
+                let meta = &t.meta;
+                let csum = t.csum;
```
matszczygiel, line 580:
```suggestion
            Some(TmpFileState{ meta, csum}) => {
```

drop-transfer/src/ws/server/v4.rs
LukasPukenis: solved

drop-transfer/src/ws/server/v6.rs
LukasPukenis: did so, thanks for the tip

matszczygiel (approved):
Nice :+1:

drop-transfer/src/ws/server/mod.rs
```diff
@@ -862,6 +863,64 @@ impl FileXferTask {
...
+
+            Some({
+                let ev = events.clone();
+                move |progress: u64| {
+                    let ev = ev.clone();
+                    async move {
+                        ev.checksum_progress(progress).await;
+                    }
+                }
+            })
```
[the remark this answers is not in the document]
LukasPukenis: that's actually very neat

matszczygiel (approved):

PR #284, opened 2024-01-23
Update moose to v4.0.1

Lipt0nas:
not sure if i should split this MR up to avoid review strain


LukasPukenis (approved):
It's a good PR, commits are squashed, the history is clean.

As a sidenote - it would be nice to add what were the changes introduced in the dependency because of a new tracker as this could give some context when debugging more easily by just swifting through the commit messages 6 months after.

In this case libmoose has fixed logging( I think ) so there should be reduced log pollution

matszczygiel (approved):

PR #289, opened 2024-02-13
Backport moose tracker update to v5.4.0

Lipt0nas:
Notable changes include:
- Context inheritance is automatic, so there is no need for manually retrieving the context from parent tracker
- Context values get updated automatically, so there is no need for periodic update task
- Function signatures had their argument order shuffled around for some reason
- We're still responsible for setting app name and version
- We no longer need to provide the version of the parent application, so this has been removed from the config

(cherry picked from commit bc25fbbd4231631250b33de34242d6d2fb4fcce2)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,37 @@
 version = 3
 
 [[package]]
-name = "ahash"
-version = "0.7.6"
+name = "addr2line"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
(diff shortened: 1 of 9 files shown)


LukasPukenis (approved):
Backported tests :+1:

matszczygiel:
The integration tests are failing due to misformatted moose events. BTW I'd change the name of the branch to `release/v5` since minor version numbers are compatible

Lipt0nas:
One analytics scenario is failing because it was pre-analytics test unflakening, it's once scenario where it expects events in a different order, locally it passes, so i guess it's not a big deal in this case

For the branch id leave this as it is, while techincally minor version versions are compatible this backport is only meant for this version specifically since others would require orders of magnitude more changes

On another note, it seems like updating the dependencies is now causing a panic in some tests? or was this a thing in the good old days?

Lipt0nas:
This version seems to be good, panics are gone, unit tests build again and analytics also passes (though still will be flaky a slight amount)

matszczygiel (approved):
+1

PR #290, opened 2024-02-13
Ensure there is no race condition in events

matszczygiel:
By ensuring there's only one connection for a given transfer at a time we prevent events from mixing with each other

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -404,22 +404,22 @@ class Wait(Action):
 class WaitAndIgnoreExcept(Action):
     def __init__(self, events: typing.List[Event]):
         self._events: typing.List[Event] = events
-        self._found: typing.List[Event] = []
 
     async def run(self, drop: ffi.Drop):
         fuse = 0
         limit = 100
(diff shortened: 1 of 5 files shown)


matszczygiel:
Waiting for in-branch testing

matszczygiel:
In-branch testing showed no signs of the bug

test/drop_test/action.py
```diff
@@ -404,22 +404,22 @@ def __str__(self):
...
+                if e in found:
                     raise Exception(f"Event {e} was received twice")
 
-                self._found.append(e)
+                found.append(e)
 
-                if len(self._found) == len(self._events):
+                if len(found) == len(self._events):
                     break
                 else:
```
LukasPukenis, line 424: This looked weird before :D

LukasPukenis (approved):

test/drop_test/action.py
matszczygiel: Yeah, and was not working with the `Repeat()` action because of the leftovers in `self._found`

PR #293, opened 2024-02-14
Add `TransferDeffered` event indicating the connection to peer couldn…

matszczygiel:
The event is emitted whenever we cannot connect to peer in a recoverable manner

### DIFF
--- a/drop-core/src/status.rs
+++ b/drop-core/src/status.rs
@@ -22,6 +22,8 @@ pub enum Status {
     FileFailed = 35,
     FileFinished = 36,
     EmptyTransfer = 37,
+    ConnectionClosedByPeer = 38,
+    TooManyRequests = 39,
 }
 
 impl serde::Serialize for Status {
--- a/drop-transfer/src/utils.rs
+++ b/drop-transfer/src/utils.rs
@@ -18,18 +18,25 @@ pub struct Hidden<T>(pub T);
 
 pub struct RetryTrigger {
     chan: watch::Receiver<()>,
-    retry: usize,
+    retry: u32,
+    retries: u32,
 }
 
 impl RetryTrigger {
-    pub fn new(chan: watch::Receiver<()>) -> Self {
-        Self { chan, retry: 0 }
+    pub fn new(chan: watch::Receiver<()>, retries: u32) -> Self {
+        Self {
+            chan,
+            retry: 0,
+            retries,
+        }
     }
 
     pub async fn backoff(&mut self) {
-        let delay = drop_config::RETRY_INTERVALS
-            .get(self.retry)
-            .map_or(Duration::MAX, |d| *d);
+        let delay = if self.retry + 1 < self.retries {
+            drop_config::FIRST_RETRY_AFTER * (0x01 << self.retry)
+        } else {
+            Duration::MAX
+        };
 
         self.retry = tokio::select! {
             _ = self.chan.changed() => 0,
--- a/drop-transfer/src/ws/client/mod.rs
+++ b/drop-transfer/src/ws/client/mod.rs
@@ -58,7 +58,7 @@ struct RunContext<'a> {
 }
 
 enum WsConnection {
-    Recoverable,
+    Recoverable(crate::Error),
     Unrecoverable(crate::Error),
     Connected(WsStream, protocol::Version),
 }
@@ -126,7 +126,18 @@ async fn connect_to_peer(
 
     let (socket, ver) = match establish_ws_conn(state, xfer, logger).await {
         WsConnection::Connected(sock, ver) => (sock, ver),
-        WsConnection::Recoverable => return ControlFlow::Continue(()),
+        WsConnection::Recoverable(error) => {
+            info!(logger, "Transfer deferred {}: {error}", xfer.id());
+
+            if state.transfer_manager.is_outgoing_alive(xfer.id()).await {
+                state.emit_event(Event::OutgoingTransferDeferred {
+                    transfer: Arc::clone(xfer),
+                    error,
+                });
+            }
+
+            return ControlFlow::Continue(());
+        }
         WsConnection::Unrecoverable(err) => {
             error!(logger, "Could not connect to peer {}: {}", xfer.id(), err);
 
@@ -198,7 +209,7 @@ async fn establish_ws_conn(
         Ok(sock) => sock,
         Err(err) => {
             debug!(logger, "Failed to connect: {:?}", err,);
-            return WsConnection::Recoverable;
+            return WsConnection::Recoverable(crate::Error::Io(err));
         }
     };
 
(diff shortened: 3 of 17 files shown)


drop-core/src/status.rs
```diff
@@ -22,6 +22,8 @@ pub enum Status {
     FileFailed = 35,
     FileFinished = 36,
     EmptyTransfer = 37,
+    UnrecognizablePeerRespone = 38,
```
LukasPukenis, line 25:
```suggestion
    UnrecognizablePeerResponse = 38,
```

test/drop_test/error.py
```diff
@@ -22,6 +22,9 @@ class Error(IntEnum):
     FILE_REJECTED = (34,)
     FILE_FAILED = (35,)
     FILE_FINISHED = (36,)
+    EMPTY_TRANSFER = (37,)
+    UNRECOGNIZABLE_PEER_RESPONE = (38,)
```
LukasPukenis, line 26: `UNRECOGNIZABLE_PEER_RESPONE` -> `UNRECOGNIZABLE_PEER_RESPONSE`

drop-transfer/src/ws/client/mod.rs
```diff
@@ -224,7 +233,7 @@ async fn establish_ws_conn(
             Ok(_) => break ver,
             Err(RequestError::General(err)) => {
                 info!(logger, "Error while making the HTTP request: {err:?}");
-                return WsConnection::Recoverable;
+                return WsConnection::Recoverable(crate::Error::UnrecognizablePeerRespone);
```
LukasPukenis, line 236: `RequestError` has `UnexpectedResponse` which by the name of it should map to `UnrecognizablePeerResponse` instead. IMO the `General` case is for the other occasions

test/scenarios.py
```diff
@@ -11295,4 +11295,40 @@
...
+                                event.File(
+                                    FILES["testfile-small"].id,
+                                    "testfile-small",
+                                    1048576,
+                                ),
+                            },
+                        )
+                    ),
+                    action.Wait(
+                        event.TransferDeferred(0, "DROP_PEER_STIMPY", Error.IO, 111)
```
LukasPukenis, line 11320: Hmm.. What does `111` stand for?

test/drop_test/event.py
```diff
@@ -417,3 +417,37 @@ def __eq__(self, rhs):
...
+
+
+class TransferDeferred(Event):
+    def __init__(
+        self,
+        uuid_slot: int,
+        peer: str,
+        status: int,
+        os_err: typing.Optional[int] = None,
+        ignore_os: bool = False,
```
LukasPukenis, line 429: I am not exactly sure what `ignore_os` does, could you explain it? It's not used in the only testcase we have

test/drop_test/ffi.py
```diff
@@ -125,7 +125,10 @@ async def wait_for_any_event(self, duration: int, ignore_progress: bool = False)
         for _ in range(0, duration):
             with self._lock:
                 self._events = [
-                    ev for ev in self._events if not isinstance(ev, event.Throttled)
+                    ev
+                    for ev in self._events
+                    if not isinstance(ev, event.Throttled)
+                    and not isinstance(ev, event.TransferDeferred)
```
LukasPukenis, line 131: I am marking this place but it refers to other places ignoring Deferred event as well - `TransferDeferred` looks pretty deterministic so we could actually not ignore it but instead alter the testcases to look for it. it should be a pretty easy change as well. What is your take on this? :)

LukasPukenis (changes requested):

LukasPukenis (changes requested):
The new addition is very nice to the usability of the library but I think the testcases mainly could be improved

drop-transfer/src/ws/client/mod.rs
matszczygiel: Yeah, true. Do you suggest creating another error type for this one?

test/scenarios.py
matszczygiel: `not connected`

test/drop_test/event.py
matszczygiel: It's the same as with `TransferFailed` event. It means that when we compare events we ignore if they differ in the OS error code

test/drop_test/ffi.py
matszczygiel: I can try not to ignore it but I fear it will blow up test cases.  The thing is that it may be emitted an unspecified amount of times before making a connection so it may break waiting for `TransferStart` events

drop-transfer/src/ws/client/mod.rs
matszczygiel: By looking into the code it doesn't make sense to change this. The only thing I see fit for a change is the name `crate::Error::UnrecognizablePeerRespone`,
The `RequestError::UnexpectedResponse` is handled cleanly in this loop, no need to report it

drop-core/src/status.rs
matszczygiel: Done

test/drop_test/error.py
matszczygiel: Done

drop-transfer/src/ws/client/mod.rs
LukasPukenis: I think there's confusion between `UnexpectedResponse` and `UnrecognizablePeerResponse`. Do you suggest to leave it as it is?

test/drop_test/ffi.py
LukasPukenis: You're totally right about this one but I have a feeling it should be fine, it doesn't retry without a delay, very curious to see if it's possible. This would result in nicer testrunner as well, those ignore parameters keep piling up :))

drop-transfer/src/ws/client/mod.rs
matszczygiel: Yeah, let me think about the naming more

matszczygiel: I've changed the name to `ConnectionClosedByPeer`

test/drop_test/ffi.py
matszczygiel: Removed it and added a configurable retry count

drop-transfer/src/utils.rs
```diff
@@ -18,18 +18,25 @@ pub struct Hidden<T>(pub T);
...
 
     pub async fn backoff(&mut self) {
-        let delay = drop_config::RETRY_INTERVALS
-            .get(self.retry)
-            .map_or(Duration::MAX, |d| *d);
+        let delay = if self.retry + 1 < self.retries {
+            drop_config::FIRST_RETRY_AFTER * (0x01 << self.retry)
+        } else {
+            Duration::MAX
+        };
```
LukasPukenis, line 39: Please make sure to add this parameter to the documentation :)

LukasPukenis:
I've tried to skim trough that commit but it's pretty complex, I agree on splitting it

matszczygiel:
@LukasPukenis I've removed the last commit from this PR

test/scenarios.py
```diff
@@ -11382,4 +11442,43 @@
...
+                                    FILES["testfile-small"].id,
+                                    "testfile-small",
+                                    1048576,
+                                ),
+                            },
+                        )
+                    ),
+                    action.Wait(
+                        event.TransferDeferred(
+                            0, "DROP_PEER_STIMPY", Error.IO, 111  # not connected
```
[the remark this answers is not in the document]
matszczygiel: Yes, it's the OS error code. The same story as with the `FileFailedEvent`

PR #294, opened 2024-02-15
Forbid emitting file events after reaching termianl state

matszczygiel:
The issue stems from throttling logic which waits for a semaphore permit. After receiving the permit it emits a `Started` event without any checks.
I've introduced the `Terminal` state in the file event emitter which drops all events emitted after reaching the terminal state.

### DIFF
--- a/drop-transfer/src/ws/events.rs
+++ b/drop-transfer/src/ws/events.rs
@@ -24,6 +24,7 @@ enum State {
     Throttled,
     Preflight,
     InFlight { started: Instant },
+    Terminal,
 }
 
 impl FileEventTxInner {
@@ -94,6 +95,11 @@ impl<T: Transfer> FileEventTx<T> {
 
     async fn start_inner(&self, events: impl IntoIterator<Item = Event>) {
         let mut lock = self.inner.lock().await;
+
+        if matches!(lock.state, State::Terminal) {
+            return;
+        }
+
         lock.state = State::InFlight {
             started: Instant::now(),
         };
@@ -111,6 +117,7 @@ impl<T: Transfer> FileEventTx<T> {
             State::Throttled => Duration::ZERO,
             State::InFlight { started } => started.elapsed(),
             State::Preflight => Duration::ZERO,
+            State::Terminal => return,
         };
 
         let phase = match event {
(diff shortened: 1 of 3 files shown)


matszczygiel:
I'm waiting for additional in-branch testing

matszczygiel:
In-branch testing confirmed the issue is fixed

drop-transfer/src/ws/events.rs
```diff
@@ -140,14 +147,15 @@ impl<T: Transfer> FileEventTx<T> {
         lock.emit_event(event);
     }
 
-    async fn force_stop(&self, event: Event, status: Result<(), i32>) {
+    async fn terminal(&self, event: Event, status: Result<(), i32>) {
```
Lipt0nas, line 150: Id maybe call this `terminate` to better convey what it does
In general curious why the rename?

drop-transfer/src/ws/events.rs
```diff
@@ -140,14 +147,15 @@ impl<T: Transfer> FileEventTx<T> {
...
         let mut lock = self.inner.lock().await;
 
-        let elapsed = match std::mem::replace(&mut lock.state, State::Idle) {
+        let elapsed = match std::mem::replace(&mut lock.state, State::Terminal) {
```
LukasPukenis, line 153: Hmmm, was this a bug in here?

drop-transfer/src/ws/events.rs
LukasPukenis: `terminal` is the term we use to describe the state so it better conveys the message about what is about to happen. `force_stop` wasn't bad but wasn't as readable :) IMO

Lipt0nas: in my mind if we already have a state `Terminal`, and we have a function that does the transition to said state, it makes sense to call it `terminate`, but this is an extremely nitpick-y and pedantic view, so feel free to just resolve :)

matszczygiel: I think I'll change the name to `terminate` since it's a verb and this is the function name so it makes sense for it to be a verb

matszczygiel: Done

drop-transfer/src/ws/events.rs
LukasPukenis: nice improvement in this area :)

LukasPukenis (commented):
approve when @Lipt0nas approves

Lipt0nas (approved):

PR #295, opened 2024-02-19
Emit transfer events from a mutex-protected closed structure

matszczygiel:
This allows for checks if the transfer is ongoing
and in case of closed transfers suppress event emission.

### DIFF
--- a/drop-transfer/src/manager.rs
+++ b/drop-transfer/src/manager.rs
@@ -23,14 +23,15 @@ use crate::{
         self,
         client::ClientReq,
         server::{FileXferTask, ServerReq},
-        FileEventTx, FileEventTxFactory, IncomingFileEventTx, OutgoingFileEventTx,
+        EventTxFactory, FileEventTx, IncomingFileEventTx, IncomingTransferEventTx,
+        OutgoingFileEventTx, OutgoingTransferEventTx, TransferEventTx,
     },
     File, FileId, FileToRecv, FileToSend, Transfer,
 };
 
 pub struct CloseResult<T: Transfer> {
-    pub xfer: Arc<T>,
-    pub events: Vec<Arc<FileEventTx<T>>>,
+    pub file_events: Vec<Arc<FileEventTx<T>>>,
+    pub xfer_events: Arc<TransferEventTx<T>>,
 }
 
 pub struct FinishResult<T: Transfer> {
@@ -62,7 +63,8 @@ pub struct IncomingState {
     pub dir_mappings: DirMapping,
     xfer_sync: sync::TransferState,
     file_sync: HashMap<FileId, IncomingLocalFileState>,
-    events: HashMap<FileId, Arc<IncomingFileEventTx>>,
+    pub file_events: HashMap<FileId, Arc<IncomingFileEventTx>>,
+    pub xfer_events: Arc<IncomingTransferEventTx>,
 }
 
 pub struct OutgoingState {
@@ -70,7 +72,8 @@ pub struct OutgoingState {
     conn: Option<UnboundedSender<ClientReq>>,
     xfer_sync: sync::TransferState,
     file_sync: HashMap<FileId, OutgoingLocalFileState>,
-    events: HashMap<FileId, Arc<OutgoingFileEventTx>>,
+    pub file_events: HashMap<FileId, Arc<OutgoingFileEventTx>>,
+    pub xfer_events: Arc<OutgoingTransferEventTx>,
 }
 
 /// Transfer manager is responsible for keeping track of all ongoing or pending
@@ -80,7 +83,7 @@ pub struct TransferManager {
     pub outgoing: Mutex<HashMap<Uuid, OutgoingState>>,
     storage: Arc<Storage>,
     logger: Logger,
-    event_factory: FileEventTxFactory,
+    event_factory: EventTxFactory,
 }
 
 #[derive(Default)]
@@ -89,7 +92,7 @@ pub struct DirMapping {
 }
 
 impl TransferManager {
-    pub fn new(storage: Arc<Storage>, event_factory: FileEventTxFactory, logger: Logger) -> Self {
+    pub fn new(storage: Arc<Storage>, event_factory: EventTxFactory, logger: Logger) -> Self {
         Self {
             incoming: Default::default(),
             outgoing: Default::default(),
@@ -99,12 +102,12 @@ impl TransferManager {
         }
     }
 
-    /// Returns `true` if the transfer is new one
+    /// Returns `Some()` if the transfer is new one
     pub async fn register_incoming(
         &self,
         xfer: Arc<IncomingTransfer>,
         conn: UnboundedSender<ServerReq>,
-    ) -> anyhow::Result<bool> {
+    ) -> anyhow::Result<Option<Arc<IncomingTransferEventTx>>> {
         let mut lock = self.incoming.lock().await;
 
         match lock.entry(xfer.id()) {
@@ -134,7 +137,7 @@ impl TransferManager {
                     }
                 }
 
-                Ok(false)
+                Ok(None)
             }
             Entry::Vacant(vacc) => {
                 if self
@@ -147,14 +150,14 @@ impl TransferManager {
                     if let Err(e) = conn.send(ServerReq::Close) {
                         warn!(self.logger, "Failed to send close request: {}", e);
                     }
-                    return Ok(false);
+                    return Ok(None);
                 }
 
                 self.storage
                     .update_transfer_sync_states(xfer.id(), sync::TransferState::Active)
                     .await;
 
-                vacc.insert(IncomingState {
+                let state = vacc.insert(IncomingState {
                     conn: Some(conn),
                     dir_mappings: Default::default(),
                     xfer_sync: sync::TransferState::Active,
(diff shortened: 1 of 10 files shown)


drop-transfer/src/ws/events.rs
```diff
@@ -404,14 +438,112 @@ impl FileEventTx<OutgoingTransfer> {
...
+        }
+
+        lock.moose.event_transfer_state(TransferStateEventData {
+            protocol_version: 0,
+            transfer_id: self.xfer.id().to_string(),
+            result: status.map_or_else(|err| err, |()| MOOSE_STATUS_SUCCESS),
+        });
+
+        lock.tx.emit(event);
+    }
```
LukasPukenis, line 468: I think this function could be simplified and `moose` could be extracted to the call site. This way it would be more uniform with other functions. And maybe even `stop()` `stop_with_moose()` `emit_ongoing()` could be squashed into the same function as their body would be the same. What do you think?

drop-transfer/src/ws/events.rs
```diff
@@ -404,14 +438,112 @@ impl FileEventTx<OutgoingTransfer> {
...
-            State::Throttled => Some(Duration::ZERO),
-            State::InFlight { started } => Some(started.elapsed()),
-            State::Preflight => Some(Duration::ZERO),
-            State::Terminal => return,
+            FileState::Idle => None,
+            FileState::Throttled => Some(Duration::ZERO),
+            FileState::InFlight { started } => Some(started.elapsed()),
+            FileState::Preflight => Some(Duration::ZERO),
+            FileState::Terminal => return,
         };
```
LukasPukenis, line 547: The renaming here makes it much more readable <3

drop-transfer/src/ws/events.rs
```diff
@@ -44,28 +36,70 @@ pub struct FileEventTx<T: Transfer> {
...
+
+trait EventTx {
+    fn emit(&self, event: Event);
+}
+
+impl EventTx for UnboundedSender<(Event, SystemTime)> {
+    fn emit(&self, event: Event) {
+        // Sometimes on shutdown it can error out. It's better not to handle this error
+        // at all
+        let _ = self.send((event, SystemTime::now()));
```
LukasPukenis, line 71: Can we at least log it as a warn? This way if we would encounter a weird situation and would have logs it might provide good insight

drop-transfer/src/manager.rs
```diff
@@ -635,16 +646,33 @@ impl TransferManager {
...
         if !lock.contains_key(&transfer_id) {
             return Err(crate::Error::BadTransfer);
         }
         self.storage.transfer_sync_clear(transfer_id).await;
-        lock.remove(&transfer_id);
+        if let Some(state) = lock.remove(&transfer_id) {
+            if !matches!(state.xfer_sync, sync::TransferState::Canceled) {
+                return Ok(Some(state.xfer_events));
+            }
+        };
```
LukasPukenis, line 664: Is it possible for this to be evaluate to `False`?

drop-transfer/src/manager.rs
```diff
@@ -635,16 +646,33 @@ impl TransferManager {
         state.ensure_not_terminated()
     }
 
-    pub async fn outgoing_remove(&self, transfer_id: Uuid) -> crate::Result<()> {
+    /// Returns `None` if the cancel event should be suppressed
```
LukasPukenis, line 649: hmm... I think this function is a bit too complex. I dislike that it knows something about the canceling and events, it should basically judging from the name - validate and remove. Maybe return what it removed.

I understand that it has multiple call sites so moving the logic into there is not scalable. Maybe you have ideas how to improve on it?

drop-transfer/src/manager.rs
matszczygiel: Yes, in case we canceled the transfer and it was not communicated to the peer yet.

drop-transfer/src/manager.rs
matszczygiel: Um, it's 13 lines of code, not that complex IMHO. It does remove the transfer if it exists and returns leftovers when appropriate.
I was thinking about emitting events here but I deliberately try to avoid emitting events from the manager because its calls are mutex-protected and I'm in fear of creating accidental deadlocks or something like that. Another reason is that I want to decouple event emissions from the manager for clarity.
However, these kinds of "hacks" are necessary to communicate to the caller what happened.
So, I sort of agree about the naming but I don't see how this can be solved better

drop-transfer/src/manager.rs
LukasPukenis: Hmm.. Could you explain what happens then? I am a bit confused about the flow here and wonder if we should somehow handle the case then or at least log it

drop-transfer/src/manager.rs
LukasPukenis: I am not particularly unhappy about amount of LoC but rather about it's responsibility, that it does something extra, like checks the state and returns an Option. If you also don't see a way to simplify it then it's fine

drop-transfer/src/ws/events.rs
matszczygiel: Um, there's no way to pass logger here sadly. We do have logs on the receiving end so that needs to suffice IMHO.
Previously it was an `.expect()` on this and I saw that it sometimes panicked in the tests. I think it was because the event callback was destroyed by Python too early, so I consider failures here to happen only on weird misuse of the event callback.

drop-transfer/src/service.rs
```diff
@@ -270,13 +271,13 @@ impl Service {
...
+                        res.file_events
+                            .iter()
+                            .map(|ev| ev.stop_silent(Status::Canceled)),
                     )
                     .await;
 
-                    self.state
-                        .emit_event(crate::Event::OutgoingTransferCanceled(res.xfer, false));
-
+                    res.xfer_events.cancel(false).await;
```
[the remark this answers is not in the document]
LukasPukenis: I am not 100% sure but I think if we would have one mutex over events(both files and transfers) it should, at least in theory simplfy and secure things :)

drop-transfer/src/ws/events.rs
matszczygiel: Sure, moose can be extracted. But `emit_ongoing()` and `stop()` cannot be squashed because the first one only checks if it's terminated and the latter set the state to `Termianted`

drop-transfer/src/manager.rs
matszczygiel: So there are two ways to cancel the transfer. Either you can do it or your peer.
1) If your peer cancels the transfer, then it will hit its `return Ok(Some(..))` line and return event tx for the calling code to emit the event
2) If you cancel the transfer, then it's not removed but its state is set to `Canceled`. This is because you still need the transfer around for communication with the peer until he acknowledges the transfer is canceled. Then you can call remove and it will not execute the code under `if !matches!()`.

There's one more option. Your peer canceled at the same time as you. Then effectively remove will be called twice and the second time it will not execute the code under `if let Some(..) = lock.remove()`

drop-transfer/src/service.rs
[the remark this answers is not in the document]
matszczygiel: I don't think it's that simple. In doing so we need to ensure all file events are flushed before canceling the transfer. Which might be very complex to ensure (remember that you can operate on files and transfer while offline).

drop-transfer/src/ws/events.rs
```diff
@@ -404,14 +438,112 @@ impl FileEventTx<OutgoingTransfer> {
     }
 }
 
+impl<T: Transfer> TransferEventTx<T> {
+    async fn emit_ongoing(&self, event: Event) {
+        let lock = self.inner.lock().await;
+
+        if let TransferState::Terminated = lock.state {
+            return;
```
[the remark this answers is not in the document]
matszczygiel: I see your point. However, I don't want to complicate the code by passing around the logger. It's fairly simple and covered by the integration tests.

I would include logs here if we would have access to some global logger.

drop-transfer/src/ws/events.rs
matszczygiel: Done

drop-transfer/src/ws/events.rs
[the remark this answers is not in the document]
LukasPukenis: Hmmm.. How painful it is to pass the logger?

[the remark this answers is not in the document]
matszczygiel: I mean it is possible of course, but we don't do this for the `FileEventTx`, so why we should do it here?

drop-transfer/src/service.rs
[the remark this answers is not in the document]
matszczygiel: So imagine that you canceled the transfer while offline, thus emitting a `TransferCanceled` event and blocking further events. Meanwhile, your peer rejected some files. So he will see the files as being rejected and you will not because you blocked these events - blocked events mean no entries in DB as well.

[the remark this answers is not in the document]
LukasPukenis: I think we misunderstood this one, I am saying that in this funciton body the file and event mutexes are separate ones and alos file mutex is locked/unlocked in each iteration, thus providing a window for a race condition. I am not saying anything about holding a mutex for whatever reason while we're offline :)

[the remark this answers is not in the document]
matszczygiel: I thought you were talking about blocking events after `TransferCanceled` not in the sense of holding mutex but in the sense of setting a `finished` flag or something.
If this is not the case, I don't see any problem with the current implementation. Nothing wrong should happen

drop-transfer/src/ws/events.rs
LukasPukenis: Fair enough

drop-transfer/src/service.rs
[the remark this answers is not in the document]
LukasPukenis: I actually don't even think the flag would be that bad. In your explained scenario with cancels and rejects, both are terminal states. And as for the mutex, I think it's still racy, why do you think nothing bad should happen?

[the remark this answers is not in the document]
matszczygiel: Events for transfer and files are not related IMHO. I gave you an example of why the flag would be bad with the offline peer. It will cause desync in what both peers state

PR #296, opened 2024-02-21
Disallow filenames and dirnames longer than 250 characters

matszczygiel:
This PR changes two things:
* Checks all dir names for being larger than the MAX length, contrary to the current check for only filenames
* Fixes a bug where too-long filenames do not block downloads and are not communicated to the peer as `Failed`

### DIFF
--- a/test/drop_test/config.py
+++ b/test/drop_test/config.py
@@ -42,8 +42,11 @@ class TestFile:
 
 
 FILES = {
-    "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456.txt": TestFile(
-        size=1 * 1024, id="p3cYEJWds51Uqb174hrbHxfE6ASz_ioErmh39jLsiWU"
+    "a" * 251
+    + ".txt": TestFile(size=1 * 1024, id="qP5G0HRpMIiygYDk_ERNe_ut5qhcDbAO2Wx8vPgXrYQ"),
+    "a" * 251
+    + "/testfile.txt": TestFile(
+        size=1 * 1024, id="NPdkvOI9aJQbptTXYYwTDvGu3aSnuR8xqAMBNe2BDgg"
     ),
     "testfile-small": TestFile(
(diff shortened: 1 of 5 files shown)


test/drop_test/config.py
```diff
@@ -45,6 +45,9 @@ def __init__(self, size: int, id: str):
     "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456.txt": TestFile(
         size=1 * 1024, id="p3cYEJWds51Uqb174hrbHxfE6ASz_ioErmh39jLsiWU"
     ),
+    "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456/testfile.txt": TestFile(
```
LukasPukenis, line 48: For such a testcase python's ability to repeat the string as `c*250` could improve it slightly

LukasPukenis: and the testcases below

LukasPukenis (dismissed):
The code is OK, but the testcases could be refactored using python's string multiplication to make it explicit about the length and also increasing readability.

test/scenarios.py
```diff
@@ -3864,10 +3856,8 @@
                                 "DROP_PEER_REN",
                                 {
                                     event.File(
-                                        FILES[
-                                            "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456.txt"
-                                        ].id,
-                                        "thisisaverylongfilenameusingonlylowercaselettersandnumbersanditcontainshugestringofnumbers01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234561234567891234567891234567890123456789012345678901234567890123456.txt",
+                                        FILES["a" * 251 + ".txt"].id,
+                                        "a" * 251 + ".txt",
```
LukasPukenis, line 3860: Looks very clean now, neat!

LukasPukenis (approved):

PR #297, opened 2024-02-23
Report file transfer error in case file subpath contains perent direc…

matszczygiel:
…tory

### DIFF
--- a/drop-transfer/src/ws/server/mod.rs
+++ b/drop-transfer/src/ws/server/mod.rs
@@ -36,7 +36,7 @@ use self::socket::{WebSocket, WsStream};
 use super::{events::FileEventTx, IncomingFileEventTx};
 use crate::{
     check,
-    file::{self, FileToRecv},
+    file::{self, FileSubPath, FileToRecv},
     protocol,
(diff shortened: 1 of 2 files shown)


drop-transfer/src/ws/server/mod.rs
```diff
@@ -1185,3 +1180,48 @@ pub fn remove_temp_files<P, I>(
 fn temp_file_name(transfer_id: uuid::Uuid, file_id: &FileId) -> String {
     format!("{}-{file_id}.dropdl-part", transfer_id.as_simple(),)
 }
+
+/// Check file and dir names are shorter then MAX and contain illegal values
+fn validate_subpath_for_download(subpath: &FileSubPath) -> crate::Result<()> {
+    const DISALLOWED: &[&str] = &[".."];
```
LukasPukenis, line 1186: Would it make sense to also add `..` in here? https://github.com/NordSecurity/libdrop/blob/0c0111fea682600e3c6a89492663310f1cd07356/drop-transfer/src/utils.rs#L109

LukasPukenis (approved):

PR #299, opened 2024-02-29
Disallow filename characters illega in FAT and EXT filesystems

matszczygiel:
As discussed in the issue ticket, we should ban characters illegal in both FAT and EXT filesystems

### DIFF
--- a/drop-transfer/src/utils.rs
+++ b/drop-transfer/src/utils.rs
@@ -105,14 +105,9 @@ pub fn filepath_variants(location: &'_ Path) -> crate::Result<impl Iterator<Item
 pub fn normalize_filename(filename: impl AsRef<str>) -> String {
     const REPLACEMENT_CHAR: &str = "_";
 
-    #[cfg(windows)]
+    // This is unified across all platforms, because we don't have a control
+    // over the filesystem mounted on user's device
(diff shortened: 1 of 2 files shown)


drop-transfer/src/utils.rs
```diff
@@ -105,15 +105,10 @@ pub fn filepath_variants(location: &'_ Path) -> crate::Result<impl Iterator<Item
...
 
-    #[cfg(all(not(windows), any(target_os = "macos", target_os = "ios")))]
-    const ILLEGAL_CHARS: &[char] = &[':', '/'];
-
-    #[cfg(all(not(windows), not(any(target_os = "macos", target_os = "ios"))))]
-    const ILLEGAL_CHARS: &[char] = &['/'];
-
     #[cfg(windows)]
     fn check_illegal_filename(mut name: String) -> String {
         const ILLEGAL: &[&str] = &[
```
LukasPukenis, line 114: That's actually a nice change of making it more uniform and simplifying the logic. I like this one a lot. Do you think it's worth adding a testcase involving all of these symbols and observing them get sanitized?

matszczygiel: We have a unit test for this (I prefer those instead of an integration test)

LukasPukenis (approved):

PR #300, opened 2024-03-04
Add a separate status for permissions error

matszczygiel:
This commit captures permissions errors from the IO errors and reports them under a separate status code

### DIFF
--- a/drop-core/src/status.rs
+++ b/drop-core/src/status.rs
@@ -24,6 +24,7 @@ pub enum Status {
     EmptyTransfer = 37,
     ConnectionClosedByPeer = 38,
     TooManyRequests = 39,
+    PermissionDenied = 40,
 }
 
 impl serde::Serialize for Status {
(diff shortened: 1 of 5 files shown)


drop-core/src/status.rs
```diff
@@ -24,6 +24,7 @@ pub enum Status {
     EmptyTransfer = 37,
     ConnectionClosedByPeer = 38,
     TooManyRequests = 39,
+    PermissionDenied = 40,
```
LukasPukenis, line 27: Just checking in - is this gonna be thrown for both sides? For example when there's insufficient pemrission to read the file and also when there aren't enough permissions to write a file on the receiver side?

matszczygiel: It's going to be thrown on the side that experienced the error. The other side receives `BadTransferState` always since we don't have a way to communicate error codes. It's the same with other error conditions as well

PR #302, opened 2024-03-05
Fix Windows bug with existing folder named as one of the transfer files

matszczygiel:


drop-transfer/src/ws/server/mod.rs
```diff
@@ -1087,6 +1087,14 @@ fn move_tmp_to_dst(
...
             }
             Err(err) => {
+                #[cfg(windows)]
+                {
+                    // On Win the permissions error is returned in case there's a directory with the
+                    // same name
+                    if path.exists() {
+                        continue;
+                    }
+                }
```
LukasPukenis, line 1097: Wondering why it doesn't trigger 
```
            Err(err) if err.kind() == io::ErrorKind::AlreadyExists => {
                continue;
            }
```
for Windows. It would be nice not to have alternate paths :thinking:

matszczygiel: Well, that's what the syscall returns. It returns `AlreadyExists` if there's a conflict with the file but if the conflict is with the folder, it's OS error 5.

matszczygiel: Well,  I presume that's the question for Microsoft. We can unify the path for all OSes, it should be no harm (I think)

LukasPukenis: Can we do that? It makes for less mental overhead when reasoning about issues of course if possible. The way I see it, it should be no harm to have this check for all OSes not differentiating windows

matszczygiel: Done

LukasPukenis: nice!

LukasPukenis (approved):

PR #303, opened 2024-03-05
Fix directories contents being merged in case their normalized name i…

matszczygiel:
…s the same

This PR fixes the simple case when two directories `</a` and `>/b` after normalization are merged into one directory `_/a` and `_/b`

### DIFF
--- a/drop-transfer/src/file/id.rs
+++ b/drop-transfer/src/file/id.rs
@@ -101,6 +101,10 @@ impl FileSubPath {
         self.0.last().expect("Missing last path component")
     }
 
+    pub fn root(&self) -> &String {
+        self.0.first().expect("Missing first path component")
+    }
+
     pub fn from_path(path: impl AsRef<Path>) -> crate::Result<Self> {
         let vec = path
             .as_ref()
--- a/drop-transfer/src/ws/server/v2.rs
+++ b/drop-transfer/src/ws/server/v2.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{hash_map::Entry, HashMap},
     fmt::Write,
     fs,
     future::Future,
@@ -26,13 +26,14 @@ use super::{
     TmpFileState,
 };
 use crate::{
+    file::FileSubPath,
     manager::FileTerminalState,
     protocol::v2,
     service::State,
     transfer::{IncomingTransfer, Transfer},
-    utils::Hidden,
+    utils::{self, Hidden},
     ws::{self, events::FileEventTx},
-    File, FileId,
+    File, FileId, FileToRecv,
 };
 
 pub struct HandlerInit<'a, const PING: bool = true> {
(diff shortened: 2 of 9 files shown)


drop-transfer/src/file/id.rs
```diff
@@ -101,6 +101,10 @@ impl FileSubPath {
         self.0.last().expect("Missing last path component")
     }
 
+    pub fn root(&self) -> &String {
+        self.0.first().expect("Missing first path component")
```
LukasPukenis, line 105: Sanity check: are we sure just a tokio threat will panic and it will return as an error?

drop-transfer/src/ws/server/v2.rs
```diff
@@ -465,6 +466,174 @@ impl handler::Downloader for Downloader {
...
+    }
+
+    let mut files = Vec::with_capacity(info.len());
+    let mut used_mappings = HashMap::new();
+
+    for (mut path, size) in info {
+        let uroot = path.root();
+        let nroot = utils::normalize_filename(uroot);
+
+        'vars: for nvariant in utils::filepath_variants(nroot.as_ref())?
```
LukasPukenis, line 506: `'vars`? Could you explain the syntax, never seen before in Rust

test/scenarios.py
```diff
@@ -4222,6 +4222,146 @@
...
+    Scenario(
+        "scenario19-5",
+        "Send two dirs for which their name is going to be the same after normalization. Expect their contents in not merged",
+        {
+            "DROP_PEER_REN": ActionList(
+                [
+                    action.Start("DROP_PEER_REN"),
+                    # Wait for another peer to appear
+                    action.WaitForAnotherPeer("DROP_PEER_STIMPY"),
+                    action.NewTransfer("DROP_PEER_STIMPY", ["/tmp/<", "/tmp/>"]),
```
LukasPukenis, line 4234: Is this correct? How do these resolve into the full paths and not error out?

LukasPukenis: Oh! these are directories

test/drop_test/config.py
```diff
@@ -137,6 +137,12 @@ def __init__(self, size: int, id: str):
     "utf8-testfile-\u5b81\u5BDF": TestFile(
         size=1 * 1024, id="iYHFSaSzwOXygOqayOaVNlF4cQUF2nyZyTzaIpnVau8"
     ),
+    "</file-01": TestFile(
+        size=1 * 1024, id="vb81tDWO3JDf_CG-MFw_pMMgOLTMxJIhmjBAVcglshk"
+    ),
+    ">/file-01": TestFile(
+        size=1 * 1024, id="d0Gycotel31QlJnnMP7hk_JKSyImdt8Fwsvp8uts2_8"
+    ),
```
LukasPukenis, line 145: I tripped on these paths because they look so unnatural, what do you think about doing something along the lines `</file-01` -> `contains-invalid-<-characters/file-01`? This way it would be in the middle and a bit more obvious

drop-transfer/src/file/id.rs
matszczygiel: It depends. The code logic forbids empty subpaths, hence this is more like an assertion

drop-transfer/src/ws/server/v2.rs
matszczygiel: https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html

drop-transfer/src/file/id.rs
LukasPukenis: Yep but I am cautious if this could result in a panic. We would handle it via our custom handler though. I can see we have more of these expects and they didn't really manifest so far so I assume it's sort of safe

drop-transfer/src/ws/server/v6.rs
```diff
@@ -667,12 +673,130 @@ impl handler::Request for (prot::TransferRequest, IpAddr, Arc<DropConfig>) {
...
+                    // New mapping, lets insert it
+                    let nroot = vacc.key().to_string();
+                    vacc.insert(uroot.to_string());
+                    nroot
+                }
+            };
+
+            let mut piter = path.iter_mut();
+            *piter.next().context("Subpath should always contain root")? = nroot;
+            piter.for_each(|s| *s = utils::normalize_filename(&*s));
```
LukasPukenis, line 713: Doesn't it call `normalize_filename` on each component of the path?

drop-transfer/src/ws/server/v6.rs
```diff
@@ -667,12 +673,130 @@ impl handler::Request for (prot::TransferRequest, IpAddr, Arc<DropConfig>) {
...
+
+            out.push(FileToRecv::new(id, path, size));
+            break 'vars;
+        }
+    }
```
LukasPukenis, line 718: I don't really see the point of using `filter_map` and then label with a `break`. Can this be done in more `rust` way or a bit simpler? The logic in this code is quite convoluted due to the problem being such so simplifying it would be beneficial

drop-transfer/src/ws/server/v6.rs
matszczygiel: It calls for all except the first one

drop-transfer/src/ws/server/v6.rs
matszczygiel: I've removed the label, but I don't see how I can simplify it further. The problem itself is not so straightforward

test/drop_test/config.py
matszczygiel: Done

drop-transfer/src/ws/server/v6.rs
LukasPukenis: Do you think it's worth renaming the function into something more generic as it works for paths instead of filenames, to make it clearer? :)

LukasPukenis (approved):

PR #304, opened 2024-03-06
Introduce file pending event

matszczygiel:
This PR introduced the `TransferPending` event

### DIFF
--- a/drop-transfer/src/service.rs
+++ b/drop-transfer/src/service.rs
@@ -184,14 +184,13 @@ impl Service {
         &mut self,
         uuid: Uuid,
         file_id: &FileId,
-        parent_dir: &Path,
+        parent_dir: &str,
     ) -> crate::Result<()> {
         debug!(
             self.logger,
-            "Client::download() called with Uuid: {}, file: {:?}, parent_dir: {}",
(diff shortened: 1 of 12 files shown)


drop-transfer/src/service.rs
```diff
@@ -200,10 +199,16 @@ impl Service {
...
+            validate_dest_path(parent_dir.as_ref())?;
+            state.file_events(file_id)?.pending(parent_dir).await;
 
             state
-                .start_download(&self.state.storage, file_id, parent_dir, &self.logger)
+                .start_download(
+                    &self.state.storage,
+                    file_id,
+                    parent_dir.as_ref(),
+                    &self.logger,
```
LukasPukenis, line 210: Why can't we do `&parent_dir` and use `.as_ref()` instead?

LukasPukenis (approved):
I like how you did not hesitate to add events to the scenarios instead of ignoring them, that's a good call and is very explicit

drop-transfer/src/service.rs
matszczygiel: Because `parent_dir` is now `str` instead of `Path`, so we need to make use of the `AsRef` trait instead of a simple reference to satisfy the `start_download()` signature

PR #305, opened 2024-03-14
Sender and receiver have mismatched file states after transfer cancellation while offline

matszczygiel:
When issuing cancel from the receiver while the sender is offline, sometimes the state of files is mismatched for peers.
The solution is to flush the changes to the sender and only then issue the cancel request

### DIFF
--- a/test/drop_test/action.py
+++ b/test/drop_test/action.py
@@ -698,6 +698,7 @@ class AssertTransfers(Action):
 
     async def run(self, drop: ffi.Drop):
         transfers = json.loads(drop.get_transfers_since(self._since_timestamp))
+        print(f"Transfers JSON:\n{json.dumps(transfers, indent=2)}")
 
         if len(transfers) != len(self._expected_outputs):
             raise Exception(
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -7460,7 +7460,7 @@ scenarios = [
     ),
     Scenario(
         "scenario29-14",
-        "Send a bunch of files and wait until some of them arrive. Then stop the sender and cancel the transfer on receiver. Expect the file states match on both sides",
+        "Send a bunch of files and wait until some of them arrive. Stop the sender and cancel the transfer on the receiver. Then, start the sender again and wait for the state to sync. Expect the file states to match on both sides",
         # TODO(msz): We should compare the results of `transfer_since()`
         # for both peers. However we don't have such posibitlity in our test
         # framework to test things across containers
--- a/test/scenarios.py
+++ b/test/scenarios.py
@@ -7459,6 +7459,178 @@ scenarios = [
         },
     ),
     Scenario(
+        "scenario29-14",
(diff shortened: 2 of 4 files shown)


matszczygiel:
Waiting for in-branch test results

test/scenarios.py
```diff
@@ -7227,6 +7227,174 @@
             ),
         },
     ),
+    Scenario(
+        "scenario29-14",
+        "Send a bunch of files and wait until some of them arrive. Then stop the sender and cancel the transfer on receiver. Expect the file states match on both sides",
```
LukasPukenis, line 7232: The description lacks the part where the seender must come online in order to sync

test/scenarios.py
```diff
@@ -7227,6 +7227,174 @@
...
+                        ]
+                    ),
+                    action.WaitForOneOf(
+                        [
+                            event.FinishFileUploaded(0, FILES["testfile-bulk-01"].id),
+                            event.FinishFileUploaded(0, FILES["testfile-bulk-02"].id),
+                            event.FinishFileUploaded(0, FILES["testfile-bulk-03"].id),
+                            event.FinishFileUploaded(0, FILES["testfile-bulk-04"].id),
+                        ]
+                    ),
```
LukasPukenis, line 7294: Wouldn't it be a bit safer for reproduction to wait for all of the files to be Uploaded and the nstop? We're not racing to do the stopping as early as possible. this way we should have a higher chance of catching the issue.

Not that anything is bad with the current setup though

LukasPukenis (approved):

test/scenarios.py
matszczygiel: That's not true. We need some of them to be `not finished` on the sender's side but finished on the receiver's. If all are finished on the sender's side then there's nothing to mismatch. That's the point

LukasPukenis: Wasn't initially clear that was the case

test/drop_test/action.py
```diff
@@ -698,6 +698,7 @@ def __init__(
 
     async def run(self, drop: ffi.Drop):
         transfers = json.loads(drop.get_transfers_since(self._since_timestamp))
+        print(f"Transfers JSON:\n{json.dumps(transfers, indent=2)}")
```
LukasPukenis, line 701: Is this useful to leave?

test/scenarios.py
matszczygiel: Done

test/drop_test/action.py
matszczygiel: Yeah, I mean it prints the DB contents. I find it more useful than not

LukasPukenis: Okey dokey

LukasPukenis (dismissed):

matszczygiel:
In-branch confirmed the issue is fixed

LukasPukenis (approved):

PR #306, opened 2024-03-14
Improve DDoS test case

matszczygiel:
It moves all test actions into the same container, thus reducing cross-container timing issues


LukasPukenis:
> It moves all test actions into the same container, thus reducing cross-container timing issues

Actually this could go into the commit message instead of PR message :)

LukasPukenis (approved):
I like the original testase as it immitates the real world better but can't deny that this must work well too

PR #308, opened 2024-03-20
Reduce severity of logs regarding transfer cleanup upon cancellation

matszczygiel:
As in the title


LukasPukenis:
It could be clearer. Judging just from the code, it appears as if it's removing legit code :)

matszczygiel:
It is removing the logs in case the transfer removal fails because it is already removed. The `warn!` logs cluttered the log output and app teams asked (I think at least twice) what's these errors. And the thing that these are no errors because it's expected that the transfer can be removed from more than one task

matszczygiel:
@LukasPukenis The thing is that if something is expected to fail and we output a log because of that it provides no value but only raises questions

LukasPukenis (approved):

PR #311, opened 2024-03-27
Introduce uniffi

matszczygiel:
This PR adds the following:
* UDL definitions of the library API (`norddrop.udl`)
* DB dump and events are strongly typed
* tests are adjusted to use on-the-fly generated Python bindings
* Rust code adjusted to work well with UDL (`uni.rs`)

I've tried not to touch the current swig-generated API hence everything in the `uni.rs` file just wraps around the ffi device and uses C-style callbacks. This should be flattened after swig removal.

The things left to do after removing the swing bindings
* convert timestamps into `SystemTime` - uniffi has direct conversions to this type
* clean the `uni.rs` file and remove the indirection to the device API (possibly removing the latter one)

So this PR is just the beginning of the changes but I want to get some feedback if things are going in the right direction.

### DIFF
--- a/norddrop/src/device/types.rs
+++ b/norddrop/src/device/types.rs
@@ -1,79 +1,44 @@
 use std::time::SystemTime;
 
-use drop_transfer::{utils::Hidden, File as _, FileId, Transfer};
+use drop_transfer::{utils::Hidden, File as _, Transfer};
 use serde::{Deserialize, Serialize};
-use uuid::Uuid;
 
-use super::utils;
-
-#[derive(Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct TransferDescriptor {
     pub path: Hidden<String>,
     pub content_uri: Option<url::Url>,
     pub fd: Option<i32>,
 }
 
-#[derive(Serialize)]
-pub struct EventTransferRequest {
-    peer: String,
-    transfer: String,
-    files: Vec<File>,
-    timestamp: u64,
+#[derive(Serialize, Deserialize)]
+pub struct File {
+    pub id: String,
+    pub path: String,
+    pub size: u64,
 }
 
-#[derive(Serialize)]
-pub struct EventRequestQueued {
-    peer: String,
-    transfer: String,
-    files: Vec<File>,
-    timestamp: u64,
-}
-
-#[derive(Serialize)]
-struct File {
-    id: String,
-    path: String,
-    size: u64,
-}
-
-#[derive(Serialize)]
-pub struct StartEvent {
-    transfer: String,
-    file: String,
-    transfered: u64,
-    timestamp: u64,
-}
-
-#[derive(Serialize)]
-pub struct ProgressEvent {
-    transfer: String,
-    file: String,
-    transfered: u64,
-    timestamp: u64,
-}
-
-#[derive(Serialize)]
+#[derive(Serialize, Deserialize)]
 pub struct Status {
-    status: u32,
+    pub status: u32,
     #[serde(skip_serializing_if = "Option::is_none")]
-    os_error_code: Option<i32>,
+    pub os_error_code: Option<i32>,
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Deserialize)]
 #[serde(tag = "reason", content = "data")]
 pub enum FinishEvent {
     TransferCanceled {
         by_peer: bool,
     },
     FileDownloaded {
-        file: String,
+        file_id: String,
         final_path: String,
     },
     FileUploaded {
-        file: String,
+        file_id: String,
     },
     FileFailed {
-        file: String,
+        file_id: String,
         #[serde(flatten)]
         status: Status,
     },
@@ -82,79 +47,100 @@ pub enum FinishEvent {
         status: Status,
     },
     FileRejected {
-        file: String,
+        file_id: String,
         by_peer: bool,
     },
 }
 
-#[derive(serde::Serialize)]
+// TODO(msz): make timestamp consistent across events
+// and make use of `SystemTime` since uniffi supports it
+#[derive(Serialize, Deserialize)]
 #[serde(tag = "type", content = "data")]
 pub enum Event {
-    RequestReceived(EventTransferRequest),
-    RequestQueued(EventRequestQueued),
-    TransferStarted(StartEvent),
-    TransferProgress(ProgressEvent),
+    RequestReceived {
+        peer: String,
+        transfer_id: String,
+        files: Vec<File>,
+        timestamp: i64,
+    },
+    RequestQueued {
+        peer: String,
+        transfer_id: String,
+        files: Vec<File>,
+        timestamp: i64,
+    },
+    TransferStarted {
+        transfer_id: String,
+        file_id: String,
+        transfered: u64,
+        timestamp: i64,
+    },
+    TransferProgress {
+        transfer_id: String,
+        file_id: String,
+        transfered: u64,
+        timestamp: i64,
+    },
     TransferFinished {
-        transfer: String,
+        transfer_id: String,
         #[serde(flatten)]
         data: FinishEvent,
-        timestamp: u64,
+        timestamp: i64,
     },
     RuntimeError {
-        status: drop_core::Status,
-        timestamp: u64,
+        status: u32,
+        timestamp: i64,
     },
     TransferPaused {
-        transfer: String,
-        file: String,
-        timestamp: u64,
+        transfer_id: String,
+        file_id: String,
+        timestamp: i64,
     },
     TransferThrottled {
-        transfer: String,
-        file: String,
+        transfer_id: String,
+        file_id: String,
         transfered: u64,
-        timestamp: u64,
+        timestamp: i64,
     },
     TransferDeferred {
-        transfer: Uuid,
+        transfer_id: String,
         peer: String,
         #[serde(flatten)]
         status: Status,
     },
     TransferPending {
-        transfer: Uuid,
-        file: FileId,
+        transfer_id: String,
+        file_id: String,
     },
-
     ChecksumStarted {
-        transfer: Uuid,
-        file: FileId,
+        transfer_id: String,
+        file_id: String,
         size: u64,
-        timestamp: u64,
+        timestamp: i64,
     },
     ChecksumFinished {
-        transfer: Uuid,
-        file: FileId,
-        timestamp: u64,
+        transfer_id: String,
+        file_id: String,
+        timestamp: i64,
     },
     ChecksumProgress {
-        transfer: Uuid,
-        file: FileId,
+        transfer_id: String,
+        file_id: String,
         bytes_checksummed: u64,
-        timestamp: u64,
+        timestamp: i64,
     },
 }
 
-#[derive(Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct Config {
-    pub dir_depth_limit: usize,
-    pub transfer_file_limit: usize,
+    pub dir_depth_limit: u64,
+    pub transfer_file_limit: u64,
     pub moose_event_path: String,
     pub moose_prod: bool,
     pub storage_path: String,
 
     #[serde(rename = "checksum_events_size_threshold_bytes")]
-    pub checksum_events_size_threshold: Option<usize>,
+    pub checksum_events_size_threshold: Option<u64>,
 
     #[serde(default = "Config::default_connection_retries")]
     pub connection_retries: u32,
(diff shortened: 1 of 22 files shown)


Lipt0nas:
I'll have a full look at this a little later, but a few points that rose in my head:
- Docs inside the UDL are probably saved for a later PR, right?
- I would advise not to convert timestamps to either of the UniFFI time types, since they might behave a little differently based on language/platform, we should either implement a custom type for this if it makes sense, or just leave it as it is

matszczygiel:
@Lipt0nas 
* correct, I want to have initial feedback so as not to waste time documenting stuff that will be changed anyway (I'll steal the documentation from the C bindings)
* can you elaborate? As far as I remember timestamps were confusing in the past, so I thought strongly typing them would solve at least some problems

uniffi-bindgen/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    uniffi::uniffi_bindgen_main();
+}
```
Lipt0nas, line 3: im not completely sure if we should keep this or not, it definitely makes development have less friction, but for CI we should be using the generator image, because otherwise we would have to keep track of the generator versions manually

Lipt0nas:
The rationale behind not using built-in time types is that there are a few minor details change based on the combination of language/platform about them (like how different languages store time a little differently (some might have more or less precision than the Rust type) c++ for example has different tick periods based on platform, though this will most likely apply to Rust as well), but the idea is that there might be subtle differences for the app teams if they would be forced to use the system types implicitly

one example i could think of that could impact us, is if we return a timestamp, this looses some information due to conversion to a native type and then they pass this back to us, for retrieval or deletion of transfers and the results are not what they expect

AFAIK the UniFFI guys are also not really recommending to use these types, so i think it might be better to just let the apps handle the timestamp as they see fit in this case

Lipt0nas (approved):
In general it looks good, i like where it's going

uniffi-bindgen/src/main.rs
matszczygiel: Yeah, that's a good point. So I overlooked the image and went the official route. With the generator image, this isn't needed but I'm creating Python bindings on the fly and it would be nice not to pull the image for running tests.
So I'm voting for leaving this, but it's not a strong opinion

matszczygiel:
@Lipt0nas you managed to convince me to use plain integers for timestamp. I'll make the changes.
Just as a side note, we were using ms in events and were expecting seconds in the `transfer_since()` call. Since I'm breaking the API I'll take the liberty to unify all timestamps to ms

uniffi-bindgen/src/main.rs
Lipt0nas: since we're not going to be mixing and matching on CI, i think this is fine to keep, since it allows us to run tests a lot easier locally

test/drop_test/event.py
```diff
@@ -27,71 +28,56 @@ def get_uuid(slot: int) -> str:
...
+def compare(s, t):
+    t = list(t)  # make a mutable copy
+    try:
+        for elem in s:
+            t.remove(elem)
+    except ValueError:
+        return False
+    return not t
+
+
```
LukasPukenis, line 40: `is_equal` could be slight improvement on naming

Lipt0nas:
Yeah, definitely agree on the unification of timestamp resolutions :+1:

test/run.py
```diff
@@ -39,7 +34,7 @@ async def main():
     if script is None:
         raise Exception("unrecognized scenario", scenario)
 
-    drop = ffi.Drop(lib, ffi.KeysCtx(runner))
+    drop = ffi.Drop(ffi.KeysCtx(runner))
```
LukasPukenis, line 37: Where is the entrypoint where I could track and see how the lib.so is being loaded and used in UniFFI generated code? Basically from my POV we don't pass the library file - how does UniFFI know where to look for it?

test/drop_test/action.py
```diff
@@ -76,6 +78,215 @@ def compare_json_struct(expected: dict, actual: dict):
...
+def compare_json_to_incoming_path_state(
+    expected: dict, actual: norddrop.IncomingPathState
+):
+    keys = ["created_at", "state"]
+
+    compare_value(expected, "incomingPathState", "created_at", actual.created_at)
+
+    if actual.kind.is_pending():
+        compare_value(expected, "IncomingPathState", "state", "pending")
+        keys.append("base_dir")
```
LukasPukenis, line 102: I do not trust this code(all these `compare_*` functions) :dagger: 
In `AssertTransfers` we encode the JSON argument - but we're typed very well - we could pass in exact typed data/objects we expect and comparison then would be free if each structure has `__hash__` defined which would be much nicer.

The testsuite is the thing which allows us to do such massive changes and trust the outputs but in this case I am not that sure. The code looks on the surface correct but the proposed method should make it more obvious and resilient imo. What do you think about this?

norddrop/src/ffi/uni.rs
```diff
@@ -0,0 +1,482 @@
...
+    ) -> Result<Self> {
+        // TODO(msz): remove indirection
+
+        let mut cb = Box::new(CbContext {
+            event_callback,
+            key_store,
+            logger,
+        });
+
+        unsafe extern "C" fn call_on_event(ctx: *mut c_void, ev: *const c_char) {
```
LukasPukenis, line 147: It looks weird to have `extern "C"` in here. I think we could get rid of all of them?

LukasPukenis (commented):
A very nice effort, very well done. Some places left me questioning so left some comments.

norddrop/src/device/types.rs
```diff
@@ -1,65 +1,30 @@
...
 pub struct Status {
-    status: u32,
+    pub status: u32,
     #[serde(skip_serializing_if = "Option::is_none")]
-    os_error_code: Option<i32>,
+    pub os_error_code: Option<i32>,
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Deserialize)]
```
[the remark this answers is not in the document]
Lipt0nas: All of the documentation about the types you can find here:
https://mozilla.github.io/uniffi-rs/Overview.html

For a slightly quicker way of glancing at what is possible to be used without any internal details, you could look at the tables here
https://mozilla.github.io/uniffi-rs/internals/lifting_and_lowering.html

norddrop/src/device/types.rs
```diff
@@ -116,45 +103,44 @@ pub enum Event {
...
+        transfer: String,
         peer: String,
         #[serde(flatten)]
         status: Status,
     },
     TransferPending {
-        transfer: Uuid,
-        file: FileId,
+        transfer: String,
+        file: String,
```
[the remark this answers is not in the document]
Lipt0nas: or alternatively we could typdef these in the bindings

test/run.py
Lipt0nas: the python bindings load the library and perform all of the setup on import

norddrop/src/device/types.rs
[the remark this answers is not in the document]
matszczygiel: @LukasPukenis I'm forwarding the uniffi calls into the `Device` methods. These use JSON. I'm doing this to not break the swig bindings for now as it would explode this PR size even bigger.
I'll clean this up in the subsequent PRs

norddrop/src/device/types.rs
```diff
@@ -87,21 +52,43 @@ pub enum FinishEvent {
     },
 }
 
-#[derive(serde::Serialize)]
+// TODO(msz): make timestamp consistent across events
+// and make use of `SystemTime` since uniffi supports it
```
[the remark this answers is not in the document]
matszczygiel: Breakage of swig API. I choose to care about this since I find it easier to split the work into multiple stages. I'll remove swig bindings in the next PR

norddrop/src/ffi/types.rs
```diff
@@ -48,6 +51,14 @@ pub enum norddrop_result {
     NORDDROP_RES_DB_ERROR = 11,
 }
 
+impl fmt::Display for norddrop_result {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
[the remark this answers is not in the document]
matszczygiel: The `fmt::Display` trait is needed for uniffy, otherwise it fails on the compilation

norddrop/src/ffi/uni.rs
```diff
@@ -0,0 +1,482 @@
...
+
+pub struct TransferState {
+    pub created_at: SystemTime,
+    pub kind: TransferStateKind,
+}
+
+pub enum IncomingPathStateKind {
+    Pending { base_dir: String },
+    Started { bytes_received: i64 },
+    Failed { status: i64, bytes_received: i64 },
```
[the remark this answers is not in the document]
matszczygiel: Good catch :+1:

test/drop_test/action.py
matszczygiel: So the thing is that we cannot use uniffy structures directly because `__eq__` compares all the fields and for example `created_at` we don't want to compare. This leaves us with two options:
* custom types and then mapping into uniffy types - this is the best approach probably, but is a ton of work (`scenarios.py` needs to be rewritten in terms of new structures)
* reuse the current JSON format and compare it with the uniffy structs - that's the thing I did in this PR

So in general I think the first option is better in the long run. I can do it but I don't want to clutter this PR even more, that's why I picked the second option for now

test/run.py
matszczygiel: I can only add that it loads the `.so` from the directory where the bindings code file is located. I've learned this the hard way

norddrop/src/ffi/mod.rs
```diff
@@ -612,7 +613,10 @@ pub extern "C" fn norddrop_get_transfers_since(
 
         let transfers = dev.transfers_since(since_timestamp)?;
 
-        Ok::<Vec<u8>, norddrop_result>(transfers.into_bytes())
+        let json = serde_json::to_string(&transfers)
+            .map_err(|_| norddrop_result::NORDDROP_RES_JSON_PARSE)?;
+
+        Ok::<Vec<u8>, norddrop_result>(json.into_bytes())
```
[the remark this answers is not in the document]
matszczygiel: Same story as before, right now I'm directing the uniffy calls into the old API layer. To be changed in the subsequent PRs

norddrop/src/ffi/uni.rs
```diff
@@ -0,0 +1,482 @@
...
+
+pub trait FdResolver: Send + Sync {
+    fn on_fd(&self, content_uri: String) -> Option<i32>;
+}
+
+pub struct NordDrop {
+    dev: Mutex<NordDropFFI>,
+    _cb: Box<CbContext>,
+    #[cfg(unix)]
+    fd_resolv: Mutex<Option<Box<FdResolverContext>>>,
```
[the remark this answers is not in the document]
matszczygiel: It must be `Mutex<Option<>>` because the `NordDrop` object must have only `fn(&self, ...)` methods. This is required by uniffy.
The double box is also needed because the C API required this `ctx: *mut c_void` to be passed into the callback. The thing is that Rust doesn't allow you to cast from `*mut c_void` to `mut dyn T`  because `Box<dyn ...>` is no longer a pointer but instead it is a [fat-pointer](https://stackoverflow.com/questions/57754901/what-is-a-fat-pointer). Hence the indirection needed to have `Box<T>` (without dyn) in the top-level structure.
Of course, it can be avoided with unsafe code, but since I'm going to remove the C API altogether I left it as it is since later on the problem would be nonexistent. Now it's just a limitation of the C API.

norddrop/src/device/types.rs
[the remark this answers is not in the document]
LukasPukenis: Greatly appreciated!

norddrop/src/ffi/uni.rs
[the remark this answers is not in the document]
matszczygiel: Done

test/drop_test/event.py
matszczygiel: Done

norddrop/src/device/types.rs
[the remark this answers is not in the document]
matszczygiel: Yeah, the `device` layer and C layer since they are somehow mixed together

test/drop_test/action.py
LukasPukenis: I totally understand it. We both agree that it needs to have concrete types but that would be a lot of work..

I think that it might be safe as we have a lot of tests and they were already passing so it should be mostly correct I suppose. Could you create a sub-ticket linked to the task to tackle this one separately after everything is done? No need for you to do that, anyone could take on.

matszczygiel: Sure thing :+1:

matszczygiel: Done

norddrop/src/ffi/uni.rs
LukasPukenis: I may have limited knowledge on this one, but `extern "C"` means "don't mangle this name as it must be ABI compatible". This is an inner function defined in `new` but it will still be not-mangled so this gives me a weird feeling - if it's exposed why have it defined inside of `new` and if it's intentional?

Also this raises a question.. Do we still need to do the whole extern "C" when we're using UniFFI?

test/run.py
matszczygiel: It does this:
```python
    if sys.platform == "darwin":
        libname = "lib{}.dylib"
    elif sys.platform.startswith("win"):
        # As of python3.8, ctypes does not seem to search $PATH when loading DLLs.
        # We could use `os.add_dll_directory` to configure the search path, but
        # it doesn't feel right to mess with application-wide settings. Let's
        # assume that the `.dll` is next to the `.py` file and load by full path.
        libname = os.path.join(
            os.path.dirname(__file__),
            "{}.dll",
        )
    else:
        # Anything else must be an ELF platform - Linux, *BSD, Solaris/illumos
        libname = "lib{}.so"

    libname = libname.format("norddrop")
    path = os.path.join(os.path.dirname(__file__), libname)
    lib = ctypes.cdll.LoadLibrary(path)
    return lib
```

norddrop/src/ffi/uni.rs
matszczygiel: The `extern "C"` specifies the C ABI calling convention and isn't related to name mangling. There's `#[no_mangle]` attribute for that.
Of course, we don't need to use it. It is the same thing as with JSON - to be removed. That's why I was not bothered with making this code nice, since it will be removed in the subsequent PRs

test/run.py
LukasPukenis: :bowing_man:

norddrop/src/ffi/uni.rs
LukasPukenis: You're right. I mistook `extern` behaviour for C and C++. Thanks.

LukasPukenis (dismissed):

norddrop/src/device/types.rs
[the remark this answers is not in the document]
matszczygiel: Done. I went the `_id` route since it involves some effort to do typedefs in the uniffy

PR #313, opened 2024-04-05
Add checksum_events_granularity config

dfetti:
The problem consists in the fact that while transferring files libdrop will generate ChecksumProgress events for every chunk of data received (256KB) which can be a lot if the file is large.

Setting the checksum_events_granularity config makes the event generation to occur at every checksum_events_granularity chunks of data.

### DIFF
--- a/drop-config/src/lib.rs
+++ b/drop-config/src/lib.rs
@@ -13,6 +13,9 @@ pub struct DropConfig {
     pub storage_path: String,
     // If set the checksum events will be emited for every file of this or bigger size
     pub checksum_events_size_threshold: Option<usize>,
+    // If set the checksum events will be emited for every checksum_events_granularity bytes
+    // Default value is 256KB.
+    pub checksum_events_granularity: u64,
     pub connection_retries: u32,
 }
 
@@ -23,6 +26,7 @@ impl Default for DropConfig {
             transfer_file_limit: 1000,
             storage_path: "libdrop.sqlite".to_string(),
             checksum_events_size_threshold: None,
+            checksum_events_granularity: 256 * 1024,
             connection_retries: 5,
         }
     }
--- a/drop-transfer/src/file/mod.rs
+++ b/drop-transfer/src/file/mod.rs
@@ -299,13 +299,14 @@ impl FileToSend {
         &self,
         limit: u64,
         progress_cb: Option<F>,
+        event_granularity: Option<u64>,
     ) -> crate::Result<[u8; 32]>
     where
         F: FnMut(u64) -> Fut + Send + Sync,
         Fut: Future<Output = ()>,
     {
         let reader = reader::open(&self.source)?.take(limit);
-        let csum = checksum(reader, progress_cb).await?;
+        let csum = checksum(reader, progress_cb, event_granularity).await?;
         Ok(csum)
     }
 }
@@ -315,6 +316,7 @@ impl FileToSend {
 pub async fn checksum<F, Fut>(
     reader: impl io::Read,
     mut progress_cb: Option<F>,
+    event_granularity: Option<u64>,
 ) -> io::Result<[u8; 32]>
 where
     F: FnMut(u64) -> Fut + Send + Sync,
@@ -325,6 +327,7 @@ where
     let mut reader = io::BufReader::with_capacity(CHECKSUM_CHUNK_SIZE, reader);
 
     let mut total_n: u64 = 0;
+    let mut from_last_event: u64 = 0;
     loop {
         let buf = reader.fill_buf()?;
         if buf.is_empty() {
@@ -337,9 +340,15 @@ where
         reader.consume(n);
 
         total_n += n as u64;
-        if let Some(progress_cb) = progress_cb.as_mut() {
-            progress_cb(total_n).await;
+
+        if let (Some(progress_cb), Some(granularity)) = (progress_cb.as_mut(), event_granularity) {
+            from_last_event += n as u64;
+            while from_last_event >= granularity {
+                progress_cb(total_n).await;
+                from_last_event -= granularity;
+            }
         }
+
         // Since these are all blocking operation we need to give tokio runtime a
         // timeslice
         tokio::task::yield_now().await;
@@ -375,6 +384,7 @@ mod tests {
         let csum = super::checksum(
             &mut &TEST[..],
             None::<fn(u64) -> futures::future::Ready<()>>,
+            None,
         )
         .await
         .unwrap();
@@ -391,7 +401,7 @@ mod tests {
 
             let size = TEST.len() as _;
             let file = super::FileToSend::from_path(tmp.path(), size).unwrap();
-            file.checksum(size, None::<fn(u64) -> futures::future::Ready<()>>)
+            file.checksum(size, None::<fn(u64) -> futures::future::Ready<()>>, None)
                 .await
                 .unwrap()
         };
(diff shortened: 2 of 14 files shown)


LukasPukenis (commented):
An integration test in scenarios would be very welcomed :)

drop-transfer/src/ws/server/mod.rs
```diff
@@ -861,7 +866,11 @@ impl FileXferTask {
             let size = tmp_size.unwrap_or(0);
             events.checksum_start(size).await;
 
-            Some(|progress| events.checksum_progress(progress))
+            Some(|progress| async move {
+                if (progress / CHECKSUM_CHUNK_SIZE as u64) % checksum_events_granularity == 0 {
```
matszczygiel, line 870: I think it might not work as expected because you cannot assume that `progress` is a multiple of `CHECKSUM_CHUNK_SIZE`. This means the logic should be more focused on accumulating progress
```rust
from_last_event += progress;
if from_last_event > granularity {
    fire_event();
    from_last_event -= granularity;
}
```

drop-config/src/lib.rs
```diff
@@ -13,6 +13,11 @@ pub struct DropConfig {
     pub storage_path: String,
     // If set the checksum events will be emited for every file of this or bigger size
     pub checksum_events_size_threshold: Option<usize>,
+    // If set to 1, we will generate progress events for every chunk (256KB)
+    // If set to 2, we will generate progress events for every 2 chunk (512KB)
+    // If set to 3, we will generate progress events for every 3 chunk (768KB), and so on..
+    // The default logic uses this as 1.
+    pub checksum_events_granularity: Option<u64>,
```
matszczygiel, line 20: The chunk size is an implementation detail. Apps should be able to specify this number in bytes.
Additionally, I'd say this shouldn't be `Option<>` and in the config JSON structure we should use
`#[serde(default =...)]`

drop-transfer/src/ws/server/mod.rs
```diff
@@ -744,10 +745,13 @@ impl FileXferTask {
...
                 events.checksum_start(self.file.size()).await;
-
                 let progress_cb = {
                     move |progress: u64| async move {
-                        events.checksum_progress(progress).await;
+                        if ((progress / CHECKSUM_CHUNK_SIZE as u64) % checksum_events_granularity)
+                            == 0
+                        {
+                            events.checksum_progress(progress).await;
+                        }
```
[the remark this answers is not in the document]
matszczygiel: Yeah, my comment was about the same thing

drop-transfer/src/ws/server/mod.rs
matszczygiel: Now I made a mistake, progress is already accumulated in `progress` so it should be something like this:
```rust
if progress - last > granularity {
    last = progress - (progress % granularity);
    fire_event();
}
```

test/scenarios.py
```diff
@@ -12317,4 +12317,192 @@
...
+                        ],
+                    ),
+                    action.CancelTransferRequest([0]),
+                    action.ExpectCancel([0], False),
+                    action.NoEvent(),
+                    action.Stop(),
+                ]
+            ),
+        },
+        tags=["moose"],
```
matszczygiel, line 12415:
```suggestion
```
This is not needed, we use tags to indicate that this test checks moose things

test/scenarios.py
```diff
@@ -12317,4 +12317,192 @@
...
```
matszczygiel, line 12506: Same

drop-transfer/src/file/mod.rs
```diff
@@ -337,9 +340,15 @@ where
...
-        if let Some(progress_cb) = progress_cb.as_mut() {
-            progress_cb(total_n).await;
+        from_last_event += n as u64;
+
+        if let (Some(progress_cb), Some(granularity)) = (progress_cb.as_mut(), event_granularity) {
+            if from_last_event >= granularity {
+                progress_cb(total_n).await;
+                from_last_event -= granularity;
+            }
         }
```
matszczygiel, line 350: This is minor since the logic is fine, but I must ask why you didn't implement the check for the event in the callback `progress_cb`? This would save you from pushing `event_granularity` down the call stack.
As I said the logic is fine and you can ignore this comment.

dfetti: Well, I had a hard time trying to store info in the progress_cb :sweat_smile:  and to avoid logic duplication there are 2 places where the callback is declared.

test/scenarios.py
```diff
@@ -12317,4 +12317,190 @@
...
+                        ],
+                    ),
+                    action.CancelTransferRequest([0]),
+                    action.ExpectCancel([0], False),
+                    action.NoEvent(),
+                    action.Stop(),
+                ]
+            ),
+        },
+    ),
```
LukasPukenis, line 12505: Good testcases. Checksumming happens when receiving file as well, meaning if A sends file to B and there's a disconnect in between then B will calculate checksum once it tries to resume the file download, so this testcase could potentially also be added.(you could model after 21-1) but it's not strictly necessary, this is already good enough :)

drop-transfer/src/file/mod.rs
```diff
@@ -337,9 +340,15 @@ where
...
 
         total_n += n as u64;
-        if let Some(progress_cb) = progress_cb.as_mut() {
-            progress_cb(total_n).await;
+        from_last_event += n as u64;
+
+        if let (Some(progress_cb), Some(granularity)) = (progress_cb.as_mut(), event_granularity) {
+            if from_last_event >= granularity {
+                progress_cb(total_n).await;
+                from_last_event -= granularity;
```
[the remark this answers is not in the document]
dfetti: We expect granularity to be bigger. I saw that, also wrote in the comment on the config field.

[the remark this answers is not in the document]
dfetti: Should it also be enforced in the code?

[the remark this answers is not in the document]
dfetti: ` // If set the checksum events will be emited for every checksum_events_granularity bytes
  // Default value is 256KB. Any value used here must be bigger than 256KB.`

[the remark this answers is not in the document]
LukasPukenis: I am coming from the defensive side - if progress would change somehow and would emit larger chunks it could break this assumption. We can't really put an `assert` in here and just crash, so I think having a loop as a failsafe is the easiest option which also allows removing that part from the comment :)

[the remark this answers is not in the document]
dfetti: Agree

test/scenarios.py
dfetti: I think it's redundant because it's the same function. Slightly different flow, but the same checksum function. Or am I missing something?

LukasPukenis: Same functionality just a different flow, yep. It's redundant in terms of functionality though. It's good enough without it, but it should be very easy to add regardless

drop-transfer/src/file/mod.rs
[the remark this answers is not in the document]
matszczygiel: I don't think this is a problem, the progress events can be missed sometimes and it does not need to be precise.
The thing is that the file is not guaranteed to be read in those 256kb chunks, it's just the buffer size given to the OS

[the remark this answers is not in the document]
LukasPukenis: I am pretty sure Libdrop emits progress of 100%(bytes==total_bytes) always, and ther'es slight risk of not emitting 100% for checksum progress

test/scenarios.py
dfetti: Ok, on it

drop-transfer/src/file/mod.rs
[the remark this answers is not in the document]
matszczygiel: You can emit manually after the `validate()` call finishes. Plus we have this `ChecksumFinised` event.
Think of it from the UI perspective. You don't need all the information, the last one relevant is enough

matszczygiel (approved):

PR #315, opened 2024-04-08
Remove SWIG generated bindings 

matszczygiel:
This PR removed things related to SWIG binding generation and added the doc comments to the UDL file.
I haven't documented the events in UDL, since I feel the structure of events should be flattened (the `FinishEvent` should be inlined in the main event structure), and I'll address this in the subsequent PR


norddrop/src/norddrop.udl
```diff
@@ -91,128 +174,325 @@ interface Event {
...
 
+	/// # Finishes the transfer from either side
+	///
+	/// # Arguments
+	/// * `transfer_id`: Transfer UUID
 	[Throws=Error]
 	void finish_transfer([ByRef] string transfer_id);
 
+	/// Removes a single transfer file from the database. The file must be rejected
+	/// beforehand, otherwise the error is returned.
```
LukasPukenis, line 450: The file must be in terminal state, not only rejected: canceled, rejected, failed

norddrop/src/norddrop.udl
```diff
@@ -91,128 +174,325 @@ interface Event {
...
+	/// # Arguments
+	/// * `peer` - Peer address.
+	/// * `descriptors` - transfer file descriptors.
+	///
+	/// # Returns
+	/// A String containing the transfer UUID.
 	[Throws=Error]
 	string new_transfer([ByRef] string peer, [ByRef] sequence<TransferDescriptor> descriptors);
 
+	/// # Finishes the transfer from either side
```
LukasPukenis, line 442: I sort of just noticed but I think `finalize` would be even better as there's no way to misinterpret it, what do you think?

LukasPukenis (dismissed):

norddrop/src/norddrop.udl
matszczygiel: Yeah, it makes sense :+1:

matszczygiel: Done

norddrop/src/norddrop.udl
matszczygiel: Done

PR #316, opened 2024-04-08
Refactor and simplify uniffi rust code after removing swig bindings

matszczygiel:
This PR removed the JSON encoding layer and simplified the `uniffy` -> `lib logic` barrier. I've changed the UDL API to be more consistent and have better names.
I've added doc comments as well


drop-transfer/src/error.rs
```diff
@@ -74,51 +74,53 @@ impl Error {
...
-            Error::AuthenticationFailed => Status::AuthenticationFailed as _,
-            Error::StorageError(_) => Status::StorageError as _,
-            Error::ChecksumMismatch => Status::FileChecksumMismatch as _,
-            Error::FileStateMismatch(FileTerminalState::Rejected) => Status::FileRejected as _,
-            Error::FileStateMismatch(FileTerminalState::Completed) => Status::FileFinished as _,
-            Error::FileStateMismatch(FileTerminalState::Failed) => Status::FileFailed as _,
-            Error::EmptyTransfer => Status::EmptyTransfer as _,
-            Error::ConnectionClosedByPeer => Status::ConnectionClosedByPeer as _,
-            Error::TooManyRequests => Status::TooManyRequests as _,
+            Error::Canceled => Status::Finalized,
```
LukasPukenis, line 82: Let's rename `Error::Canceled` -> `Finalized` as well

norddrop/src/device.rs
```diff
@@ -528,39 +477,39 @@ impl NordDropFFI {
         }
 
         for desc in descriptors {
-            if let Some(content_uri) = &desc.content_uri {
-                #[cfg(target_os = "windows")]
-                {
-                    let _ = content_uri;
+            match desc {
```
LukasPukenis, line 480: Just checking in that `cfg(target_os = "windows")` is missing, is that intended?

norddrop/src/types.rs
```diff
@@ -0,0 +1,97 @@
...
+        impl From<$to> for $from {
+            fn from(t: $to) -> $from {
+                match t {
+                    $($to::$t => $from::$f),+
+                }
+            }
+        }
+    };
+}
+
```
LukasPukenis, line 88: Do we still need these?

LukasPukenis (approved):
Nice

norddrop/src/device.rs
matszczygiel: There's `cfg(unix)` instead, It's (almost) the same thing but reversed and more concise

norddrop/src/types.rs
matszczygiel: Well, let me check, but probably we do

matszczygiel: Yeah, we do. This is the translation of `slog` levels to `our` levels needed for the logging infrastructure

drop-transfer/src/error.rs
matszczygiel: Huh, now as I think of it, it doesn't make sense for the error, because it means that something is canceled. For example, this error is handled cleanly (without emitting an event) in many cases.
So I see this as a different entity from `Finalization` and I think we should leave it as it is

LukasPukenis: Thanks for explaining. Where is it used though? In a context of transfer or a file?

matszczygiel: In the context of file, for example, it's used to signal disconnection from the WS layer to the logic above (`Paused` event emitted)

LukasPukenis: Thanks for explaining, now it's clear. It feels like this enum value should hold more information why it was cancelled or else we're just guessing, wonder if that doesn't hide any bugs.. It was always there, just is now highlighted due to a rename.

By the way
```
...
elif ev.is_transfer_failed():
        return event.FinishFailedTransfer(
            transfer_slot, ev.status.status, ev.status.os_error_code
        )
    elif ev.is_transfer_deferred():
        return event.TransferDeferred(
            transfer_slot, ev.peer, ev.status.status, ev.status.os_error_code
        )
...
```

These functions are generated by UniFFI I assume? How do they work, does UniFFI generate `is_*` itself or do we have something tied to it somewhere(can;t find it)?

LukasPukenis: That's an additional point for commiting generated code

PR #321, opened 2024-04-12
Fix checksum progress for small granularity

dfetti:
There was a problem in the previous version if the checksum progress granularity is set to a value smaller than 256KB. The progress bytes were not reported correctly. This commit fixes it.

Also make the last checksum progress event to report the file size.

And also add more test scenarios.

### DIFF
--- a/drop-transfer/src/file/mod.rs
+++ b/drop-transfer/src/file/mod.rs
@@ -327,10 +327,19 @@ where
     let mut reader = io::BufReader::with_capacity(CHECKSUM_CHUNK_SIZE, reader);
 
     let mut total_n: u64 = 0;
-    let mut from_last_event: u64 = 0;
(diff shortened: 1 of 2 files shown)


drop-transfer/src/file/mod.rs
```diff
@@ -327,10 +327,19 @@ where
...
         if buf.is_empty() {
+            // If we reached the end of file and the already announced_bytes are different
+            // than the total file size, we announce the total file size
+            // It simply means that the file size is not a multiple of granularity
+            if let Some(progress_cb) = progress_cb.as_mut() {
+                if announced_bytes != total_n {
+                    progress_cb(total_n).await;
+                }
+            }
+
```
LukasPukenis, line 342: This one is a nice improvement

PR #323, opened 2024-04-16
Split checksum events into finalize and verify

dfetti:
Until now, the API users had to rely on their own state machine to make the difference between finalizing and verifying checksum events.

Now we generate different events for those cases FinalizeChecksum*/VerifyChecksum*.


LukasPukenis (commented):
Nice implementation.

One addition I would like to have is at least one testcase checking all 3 verify events, such as scenario53-8 has with finalize checksumming events

matszczygiel:
How did these changes in `Cargo.lock` appear?

LukasPukenis (commented):
+1

dfetti:
About the Cargo.lock, I think they are leftovers from old ffi removal.

matszczygiel (approved):

PR #324, opened 2024-04-19
Libdrop is dynamic lib everywhere

LukasPukenis:


LukasPukenis:
This should be fine, the pipeline in libdrop-build passes. Merging

PR #327, opened 2024-04-26
Include `base_dir` field in the `RequestQueued` event files

matszczygiel:
The title says it all


LukasPukenis (approved):
nice and quick one

PR #332, opened 2024-05-21
FILE-601 Receiver does not reports events

matszczygiel:
All the server's `HandlerLoop` implementations were firing `FileEventTx::pause()` in a spawned tokio task. I assume that spawning tokio task took a lot of time for a transfer with multiple big files and a new connection was started before it finished. Thus blocking events for the next transfer.


drop-transfer/src/ws/server/v4.rs
```diff
@@ -476,20 +488,18 @@ impl handler::HandlerLoop for HandlerLoop<'_> {
                 .map(|tmp| (tmp.base_path, FileId::from(tmp.file_id))),
         );
     }
+
+    // While the destructur ensures the events are paused this function waits for
```
LukasPukenis, line 492:
```suggestion
    // While the destructor ensures the events are paused this function waits for
```

drop-transfer/src/ws/server/v6.rs
```diff
@@ -370,6 +370,18 @@ impl HandlerLoop<'_> {
             }
         }
     }
+
+    fn take_pause_futurues(&mut self) -> impl Future<Output = ()> {
```
LukasPukenis, line 374:
```suggestion
    fn take_pause_futures(&mut self) -> impl Future<Output = ()> {
```

But also the name is pretty cryptic, it's more like `fn finish_jobs()`

test/scenarios.py
```diff
@@ -2664,6 +2664,204 @@
             ),
         },
     ),
+    Scenario(
+        "scenario11-4",
+        "Initate a transfers with multiple. Wait for then to start and then stop the sender. Then restart sender and expect transfer to finish successfuly",
```
LukasPukenis, line 2669:
```suggestion
        "Initate a transfers with multiple files. Wait for them to start and then stop the sender. Then restart sender and expect transfer to finish successfuly",
```

test/scenarios.py
```diff
@@ -2664,6 +2664,204 @@
...
+                            norddrop.QueuedFile(FILES["testfile-bulk-05"].id, "testfile-bulk-05", 10485760, "/tmp"),
+                            norddrop.QueuedFile(FILES["testfile-bulk-06"].id, "testfile-bulk-06", 10485760, "/tmp"),
+                            norddrop.QueuedFile(FILES["testfile-bulk-07"].id, "testfile-bulk-07", 10485760, "/tmp"),
+                            norddrop.QueuedFile(FILES["testfile-bulk-08"].id, "testfile-bulk-08", 10485760, "/tmp"),
+                            norddrop.QueuedFile(FILES["testfile-bulk-09"].id, "testfile-bulk-09", 10485760, "/tmp"),
+                            norddrop.QueuedFile(FILES["testfile-bulk-10"].id, "testfile-bulk-10", 10485760, "/tmp"),
+                        ]),
+                    ),
+                    # fmt: on
+                    # fmt: off
```
LukasPukenis, line 2706: hmm?

test/scenarios.py
```diff
@@ -2664,6 +2664,204 @@
...
+                        event.Start(0, FILES["testfile-bulk-05"].id),
+                        event.Start(0, FILES["testfile-bulk-06"].id),
+                        event.Start(0, FILES["testfile-bulk-07"].id),
+                        event.Start(0, FILES["testfile-bulk-08"].id),
+                        event.Start(0, FILES["testfile-bulk-09"].id),
+                        event.Start(0, FILES["testfile-bulk-10"].id),
+                    ])], 4),
+                    # fmt: on
+                    action.Stop(),
+                    # fmt: off
```
LukasPukenis, line 2722: are these #fmt comments all necessary?

test/scenarios.py
```diff
@@ -2664,6 +2664,204 @@
...
+                        event.Paused(0, FILES["testfile-bulk-05"].id),
+                        event.Paused(0, FILES["testfile-bulk-06"].id),
+                        event.Paused(0, FILES["testfile-bulk-07"].id),
+                        event.Paused(0, FILES["testfile-bulk-08"].id),
+                        event.Paused(0, FILES["testfile-bulk-09"].id),
+                        event.Paused(0, FILES["testfile-bulk-10"].id),
+                    ]),
+                    # fmt: on
+                    action.Sleep(2),
+                    action.ClearEventQueue(),
```
LukasPukenis, line 2738: Curious, what does this achieve?

drop-transfer/src/ws/server/v6.rs
matszczygiel: Well, I meant to take out the future with the closing task. Hence this function is not `async` but returns owned future

test/scenarios.py
matszczygiel: Um, they are not but the effect of the formatted on those large arrays of events is really destructive and it's hard to read

test/scenarios.py
matszczygiel: So, there might be some leftover events like `paused` or `started`. We don't know how many of them to expect since the `Stop()` can happen pretty independently of the receiver sending requests.

drop-transfer/src/ws/server/v4.rs
matszczygiel: Done

drop-transfer/src/ws/server/v6.rs
matszczygiel: Typo fixed

LukasPukenis (approved):

PR #335, opened 2024-07-10
Remove protocols V1, V2, V4 and V5

matszczygiel:
I am removing the V1 and V2 protocols. Should I remove protocols V4 and V5 as well in this PR?


LukasPukenis:
Is there a reason not to remove? The reason why we remove is because of missing authorization and v5 has partial authorization, I think if we're heading down this path it's worth to remove those extra versions as well :) unless I am missing somethink?

LukasPukenis (dismissed):

Lipt0nas (approved):
Id like to see more diffs like these :D

PR #337, opened 2024-07-10
Implement automatic transfer cancelation

matszczygiel:
This PR adds the automatic cancelation feature. This is done by checking on each file reaching any terminal state to see if there's any file not yet in the terminal state. If there isn't, the transfer is automatically canceled by the peer who receives the notification about the file state change.
Additionally, I've implemented a check upon establishing the connection to see if the transfer should be canceled based on file states to account for transfers created by old libdrop versions and to be double sure the mechanism works.

### DIFF
--- a/drop-transfer/examples/udrop.rs
+++ b/drop-transfer/examples/udrop.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::{btree_map::Entry, BTreeMap, HashSet},
     env,
     io::Write,
     net::IpAddr,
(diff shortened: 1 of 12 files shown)


drop-transfer/src/manager.rs
```diff
@@ -35,8 +35,30 @@ pub struct CloseResult<T: Transfer> {
 }
 
 pub struct FinishResult<T: Transfer> {
-    pub xfer: Arc<T>,
-    pub events: Arc<FileEventTx<T>>,
+    pub xfer_state: FinishTrasferState<T>,
+    pub file_events: Arc<FileEventTx<T>>,
+}
+
+pub enum FinishTrasferState<T: Transfer> {
```
Lipt0nas, line 42:
```suggestion
pub enum FinishTransferState<T: Transfer> {
```

drop-transfer/examples/udrop.rs
```diff
@@ -202,93 +201,19 @@ async fn listen(
...
-                active_file_downloads.remove(&xfer.id());
+        if let Event::RequestReceived(xfer) = ev {
+            let xfid = xfer.id();
+            for file in xfer.files().values() {
+                service
+                    .download(xfid, file.id(), &out_dir.to_string_lossy())
+                    .await
+                    .context("Cannot issue download call")?;
             }
-            _ => (),
```
Lipt0nas, line 291: this takes A LOT of state management logic out, hopefully the apps will get similar gains :tada:

Lipt0nas (approved):

drop-transfer/examples/udrop.rs
matszczygiel: Yeah, removing this code was quite motivating to make this feature

drop-transfer/src/manager.rs
matszczygiel: Done

Lipt0nas (approved):

PR #338, opened 2024-07-12
Validate file ID for path traversal

matszczygiel:


changelog.md
```diff
@@ -2,6 +2,7 @@
 ### **Fresh air**
 ---
 * Remove support for deprecated, unsecure protocols V1, V2, V4 and V5
+* Sanitaze file ID before downloading a file, to prevent the temporary file from being allowed to traverse parent dirs
```
Lipt0nas, line 5:
```suggestion
* Sanitize file ID before downloading a file, to prevent the temporary file from being allowed to traverse parent dirs
```

drop-transfer/src/ws/server/mod.rs
```diff
@@ -1164,9 +1165,24 @@ fn validate_subpath_for_download(subpath: &FileSubPath) -> crate::Result<()> {
     Ok(())
 }
 
+/// Check file ID for illegal characters so that the temp file is not created in
+/// parent directories
+fn validate_file_id_for_download(file_id: &FileId) -> crate::Result<()> {
+    // Ideally we would allow only alphanumeric characters but we use URL base64
+    // which allows for more
+    const DISALLOWED: &[char] = &['.', '/', '\\', ':'];
```
Lipt0nas, line 1173: since Base64 consists of alphanumeric chars + 3, maybe it would be better to have a whitelist as opposed to a blacklist for chars?

changelog.md
matszczygiel: Good catch :+1:

drop-transfer/src/ws/server/mod.rs
matszczygiel: Yeah, that would make sense as well. Lemme change this accordingly

changelog.md
matszczygiel: Done

drop-transfer/src/ws/server/mod.rs
matszczygiel: Done

Lipt0nas (approved):

PR #340, opened 2024-07-29
FILE-613: retrieve private key always

LukasPukenis:
Private key retrieval was previously done on `device::new()` once and stored, however API users sometimes hcange the private key and need to recreate libdrop instance. However it's no tthe ideal way of doing things and thus this PR just queries for the private key each time it needs.

### DIFF
--- a/norddrop/src/device.rs
+++ b/norddrop/src/device.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use drop_analytics::DeveloperExceptionEventData;
-use drop_auth::{PublicKey, SecretKey};
+use drop_auth::{PublicKey, SecretKey, PUBLIC_KEY_LENGTH, SECRET_KEY_LENGTH};
 use drop_config::{Config, DropConfig, MooseConfig};
 use drop_storage::types::Transfer as TransferInfo;
 use drop_transfer::{auth, utils::Hidden, Event, FileToSend, OutgoingTransfer, Service, Transfer};
@@ -15,7 +15,7 @@ use tokio::{
     task::JoinHandle,
 };
 
-use crate::{event, TransferDescriptor};
+use crate::{event, KeyStore, TransferDescriptor};
 
 pub type Result<T = ()> = std::result::Result<T, crate::LibdropError>;
 
@@ -52,14 +52,10 @@ impl EventDispatcher {
 impl NordDropFFI {
     pub(super) fn new(
         event_cb: impl Fn(crate::Event) + Send + Sync + 'static,
-        pubkey_cb: impl Fn(IpAddr) -> Option<PublicKey> + Send + 'static,
-        privkey: SecretKey,
+        key_store: Arc<dyn KeyStore>,
         logger: Logger,
     ) -> Result<Self> {
         trace!(logger, "norddrop_new()");
-
-        // It's a debug print. Not visible in the production build
-        debug!(logger, "Private key: {:02X?}", privkey.to_bytes());
 
         Ok(NordDropFFI {
             instance: Arc::default(),
@@ -69,7 +65,7 @@ impl NordDropFFI {
                 cb: Arc::new(event_cb) as _,
             },
             config: DropConfig::default(),
-            keys: Arc::new(crate_key_context(logger, privkey, pubkey_cb)),
+            keys: Arc::new(create_key_context(logger, key_store)),
             #[cfg(unix)]
             fdresolv: None,
         })
@@ -522,22 +518,33 @@ impl NordDropFFI {
     }
 }
 
-fn crate_key_context(
-    logger: slog::Logger,
-    privkey: SecretKey,
-    pubkey_cb: impl Fn(IpAddr) -> Option<PublicKey> + Send + 'static,
-) -> auth::Context {
-    let pubkey_cb = std::sync::Mutex::new(pubkey_cb);
-    let public = move |ip: IpAddr| {
-        let guard = pubkey_cb.lock().expect("Failed to lock pubkey callback");
-        let key = guard(ip)?;
-        drop(guard);
+fn create_key_context(logger: slog::Logger, key_store: Arc<dyn KeyStore>) -> auth::Context {
+    let privkey = {
+        let key_store = key_store.clone();
+        let logger = logger.clone();
+        let privkey_cb = std::sync::Mutex::new(key_store);
+        move || {
+            let guard = privkey_cb.lock().expect("Failed to lock privkey callback");
+            let privkey: [u8; SECRET_KEY_LENGTH] = guard.privkey().try_into().ok()?;
+            drop(guard);
 
-        debug!(logger, "Public key for {ip:?}: {key:?}");
-        Some(key)
+            debug!(logger, "Retrieved private key: {:?}", privkey);
+            Some(SecretKey::from(privkey))
+        }
     };
 
-    auth::Context::new(privkey, public)
+    let pubkey_cb = std::sync::Mutex::new(key_store);
+    let pubkey = move |ip: IpAddr| {
+        let guard = pubkey_cb.lock().expect("Failed to lock pubkey callback");
+        let pubkey = guard.on_pubkey(ip.to_string())?;
+        drop(guard);
+
+        let pubkey: [u8; PUBLIC_KEY_LENGTH] = pubkey.try_into().ok()?;
+        debug!(logger, "Retrieved public key for: {} key: {:?}", ip, pubkey);
+        Some(PublicKey::from(pubkey))
+    };
+
+    auth::Context::new(privkey, pubkey)
 }
 
 fn open_database(
--- a/norddrop/src/uni.rs
+++ b/norddrop/src/uni.rs
@@ -1,7 +1,5 @@
 use std::sync::Mutex;
 
-use drop_auth::{PublicKey, SecretKey, PUBLIC_KEY_LENGTH};
-
 use crate::{device::NordDropFFI, Event, TransferDescriptor, TransferInfo};
 
 pub type Result<T> = std::result::Result<T, crate::LibdropError>;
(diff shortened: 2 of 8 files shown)


norddrop/src/device.rs
```diff
@@ -522,22 +522,29 @@ impl NordDropFFI {
...
-    logger: slog::Logger,
-    privkey: SecretKey,
-    pubkey_cb: impl Fn(IpAddr) -> Option<PublicKey> + Send + 'static,
-) -> auth::Context {
-    let pubkey_cb = std::sync::Mutex::new(pubkey_cb);
-    let public = move |ip: IpAddr| {
-        let guard = pubkey_cb.lock().expect("Failed to lock pubkey callback");
-        let key = guard(ip)?;
-        drop(guard);
+fn crate_key_context(logger: slog::Logger, key_store: Arc<Box<dyn KeyStore>>) -> auth::Context {
```
Lipt0nas, line 525:
```suggestion
fn create_key_context(logger: slog::Logger, key_store: Arc<Box<dyn KeyStore>>) -> auth::Context {
```

norddrop/src/device.rs
```diff
@@ -522,22 +522,29 @@ impl NordDropFFI {
...
+        let key_store = Arc::clone(&priv_keystore);
+
+        let privkey: [u8; PUBLIC_KEY_LENGTH] = key_store.privkey().try_into().ok()?;
+        debug!(priv_logger, "Retrieved private key: {:?}", privkey);
+        Some(SecretKey::from(privkey))
+    };
+
+    let pub_keystore = Arc::clone(&key_store);
+    let pubkey = move |ip: IpAddr| {
+        let key_store = Arc::clone(&pub_keystore);
```
Lipt0nas, line 539:
```suggestion
    let priv_keystore = key_store.clone();
    let priv_logger = logger.clone();
    let privkey = move || {
        let key_store = priv_keystore.clone();

        let privkey: [u8; PUBLIC_KEY_LENGTH] = key_store.privkey().try_into().ok()?;
        debug!(priv_logger, "Retrieved private key: {:?}", privkey);
        Some(SecretKey::from(privkey))
    };

    let pubkey = move |ip: IpAddr| {
        let key_store = key_store.clone();
```

Can get rid of one clone here, and less clutter with using straight `.clone()`

norddrop/src/device.rs
```diff
@@ -522,22 +522,29 @@ impl NordDropFFI {
...
-        debug!(logger, "Public key for {ip:?}: {key:?}");
-        Some(key)
+        let pubkey = key_store.on_pubkey(ip.to_string())?;
+        let pubkey: [u8; PUBLIC_KEY_LENGTH] = pubkey.try_into().ok()?;
+        debug!(logger, "Retrieved public key for: {} key: {:?}", ip, pubkey);
+        Some(PublicKey::from(pubkey))
     };
 
-    auth::Context::new(privkey, public)
+    auth::Context::new(privkey, pubkey)
```
Lipt0nas, line 547: Is this fine in general? the previous version used a mutex here, is it okay to remove it here?

LukasPukenis: I was thinking about this mutex before removal quite hard and I can only see one reason for it being there - to serialize libdrop's callbacks, meaning if two would happen simultaneously, they would go one after another. I don't see any reason for this being a problem at all to be honest.

As per mutex - it should be implemented in KeyStore implementation on the app side that actually controls the keys, it should then lock the mutex on the callback and when updating it. Guarding with mutex in libdrop just serializes the callback invocations for no purpose IMO. @matszczygiel might have an opinion on this

norddrop/src/device.rs
LukasPukenis: hehe

norddrop/src/device.rs
```diff
@@ -522,22 +521,30 @@ impl NordDropFFI {
...
+        move || {
+            let privkey: [u8; PUBLIC_KEY_LENGTH] = key_store.privkey().try_into().ok()?;
+            debug!(logger, "Retrieved private key: {:?}", privkey);
+            Some(SecretKey::from(privkey))
+        }
+    };
+
+    let pubkey = {
+        move |ip: IpAddr| {
+            let key_store: Arc<Box<dyn KeyStore>> = Arc::clone(&key_store);
```
Lipt0nas, line 538: mega nit: mixing `Arc::clone` and `.clone()`

Lipt0nas (approved):
Looks good from my side, don't know if we want to wait for Mateusz to return from vacation about the mutex question

norddrop/src/device.rs
LukasPukenis: oh didn't notice, he's on vacation.

matszczygiel: Hey, IMHO Mutex should be there since we cannot trust apps to provide thread-safe callback

matszczygiel: Imagine app fetching the key from hash map in the callback and at the same time updating the hash map on some other thread with new keys or more peers

norddrop/src/device.rs
```diff
@@ -52,15 +55,11 @@ impl EventDispatcher {
 impl NordDropFFI {
     pub(super) fn new(
         event_cb: impl Fn(crate::Event) + Send + Sync + 'static,
-        pubkey_cb: impl Fn(IpAddr) -> Option<PublicKey> + Send + 'static,
-        privkey: SecretKey,
+        key_store: Arc<Box<dyn KeyStore>>,
```
matszczygiel, line 58: Does `Arc<dyn KeyStore>` work? Should be, no need for double indirection

norddrop/src/device.rs
```diff
@@ -522,22 +521,30 @@ impl NordDropFFI {
...
-    logger: slog::Logger,
-    privkey: SecretKey,
-    pubkey_cb: impl Fn(IpAddr) -> Option<PublicKey> + Send + 'static,
-) -> auth::Context {
-    let pubkey_cb = std::sync::Mutex::new(pubkey_cb);
-    let public = move |ip: IpAddr| {
-        let guard = pubkey_cb.lock().expect("Failed to lock pubkey callback");
-        let key = guard(ip)?;
-        drop(guard);
+fn create_key_context(logger: slog::Logger, key_store: Arc<Box<dyn KeyStore>>) -> auth::Context {
```
matszczygiel, line 524: Same here

norddrop/src/device.rs
matszczygiel: IMHO, this clone is not needed

norddrop/src/uni.rs
```diff
@@ -36,20 +34,9 @@ impl NordDrop {
...
-
         let dev = NordDropFFI::new(
             move |ev| event_callback.on_event(ev),
-            move |peer_ip| {
-                let pubkey = key_store.on_pubkey(peer_ip.to_string())?;
-                let pubkey: [u8; PUBLIC_KEY_LENGTH] = pubkey.try_into().ok()?;
-                Some(PublicKey::from(pubkey))
-            },
-            privkey,
+            Arc::new(key_store),
```
matszczygiel, line 39: You can convert `Box<T>` to `Arc<T>` by `.into()`

norddrop/src/device.rs
LukasPukenis: But the problem is that it doesn't prevent the race, we do not lock the mutex when there's an update of the value(as we don't modify ourselves), only when reading, thus we basically just prevent libdrop from reading in parallel but don't guard against the modifications unless I miss something?

matszczygiel: Yes, we cannot fully protect from apps doing data races, but at least we can defend as much as we can like the callback is modifying something (on the app side) without mutex protection. We cannot guarantee this doesn't happen

LukasPukenis: Mutex only makes sense when both readers and writers are aware of it and block on locking, thus if the app is not aware of our internal mutex, I don't see the benefit we get. I would speak differently if libdrop would also modify the keys in the callback, in that case our mutex at least would guard to some degree IMO

matszczygiel: The example where the app is modifying a data structure in the callback, and the mutex is protecting us from data race
```cpp
class Callback: KeyStore {
    private:
        Cache _cache;
        DB _db;
    public:
        vector<char> on_pubkey(const char* peer) override {
            if _cache.contains(peer) {
                return _cache.key(peer);
            }

            auto key = _db.fetch(peer);
            _cache.insert(peer, key); // <-- this is writing to some memory structure
            return key;
        }
}
```

LukasPukenis: Restored the Mutex on Public-Key retrieval. Skipped it on the Private-Key retrieval

norddrop/src/device.rs
LukasPukenis: Apparently yes

norddrop/src/device.rs
```diff
@@ -15,13 +15,16 @@ use tokio::{
...
-use crate::{event, TransferDescriptor};
+use crate::{event, KeyStore, TransferDescriptor};
 
 pub type Result<T = ()> = std::result::Result<T, crate::LibdropError>;
 
 const SQLITE_TIMESTAMP_MIN: i64 = -210866760000;
 const SQLITE_TIMESTAMP_MAX: i64 = 253402300799;
 
+pub const PUBLIC_KEY_LENGTH: usize = 32;
+pub const SECRET_KEY_LENGTH: usize = 32;
```
matszczygiel, line 26: Those constants are available in `drop-auth/src/lib.rs` and you can use those. Defining those in two places can be error-prone

LukasPukenis:
@matszczygiel addressed the comments

matszczygiel (dismissed):
Looks good, however, I am not sure about dropping the mutex around `privkey()`

matszczygiel (approved):

LukasPukenis:
All comments addressed @Lipt0nas @matszczygiel . I gave a pipeline build to Windows to test. The tests failing in natlab are due to flakyness

Lipt0nas (approved):

PR #353, opened 2025-02-28
Remove the migration for incoming_path_pending_states

LukasPukenis:
See commit message

### DIFF
--- a/drop-storage/migrations/003-retries/up.sql
+++ b/drop-storage/migrations/003-retries/up.sql
@@ -80,11 +80,25 @@ DROP TABLE outgoing_path_pending_states;
 
 ALTER TABLE incoming_path_pending_states ADD COLUMN base_dir TEXT NOT NULL DEFAULT '';
 
-UPDATE incoming_path_pending_states SET base_dir = ipss.base_dir
-FROM (SELECT path_id, base_dir, MIN(created_at) FROM incoming_path_started_states GROUP BY path_id) as ipss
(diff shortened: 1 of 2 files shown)


drop-storage/migrations/003-retries/up.sql
```diff
@@ -80,11 +80,25 @@ DROP TABLE outgoing_path_pending_states;
...
+CREATE TABLE IF NOT EXISTS incoming_path_started_states_new (
+  id INTEGER PRIMARY KEY AUTOINCREMENT, 
+  path_id INTEGER NOT NULL,
+  bytes_received INTEGER NOT NULL,
+  created_at TIMESTAMP NOT NULL DEFAULT(STRFTIME('%Y-%m-%d %H:%M:%f', 'NOW')),
+  FOREIGN KEY(path_id) REFERENCES incoming_paths(id) ON DELETE CASCADE ON UPDATE CASCADE,
+  CHECK(bytes_received >= 0)
+);
+
+INSERT INTO incoming_path_started_states_new (id, path_id, bytes_received, created_at)
```
matszczygiel, line 95: Have you double-checked that inserting `id` doesn't somehow break `AUTOINCREMENT`?

matszczygiel: Ok, I've checked it, doesn't seem like. All good then

matszczygiel (approved):

